PR #142, opened 2021-08-30
Adding updated ES data.

marklise:


siva-ka:
captured in https://github.com/bcgov/bcparks.ca/pull/158

PR #163, opened 2021-09-14
BC-487 DevOps Update (CMS)

BcGovNeal:
# Description

Build and deployment updates to the `CMS` and `PostgreSQL` components
https://bcdevex.atlassian.net/browse/BC-487

This PR includes the following proposed change(s):

**CMS Changes**

- Added Helm Chart for BCParks.  Only `CMS` and `PostgreSQL` chart componentss for now.  Others will follow
- Modified Docker build to use `s2i` builds https://github.com/openshift/source-to-image
- Modified Strapi code to use environment variables directly, this allows image reuse
  - `server.js` is no longer generated at container start.  It will now pull values from container/os environment variables
  - Lowered default DB pool `min` and `max` connections to `2` and `10` respectively.  Previous values were too high and used up Postgres connections too fast
  - **IMPORTANT**: Many of the environment variables have been prefixed with `STRAPI_`.  Please use `.env.example` file as reference and update your local `.env` file accordingly
- **NEW** CMS will now defaults to using `sqlite`.  Please set `DATABASE_CLIENT=postgres` if you want to keep using postgres locally 

**Github Action Changes**

- The old `main-build.yaml` GA workflow has been removed.  Workflows for each component will be added separately
- CMS DEV build and deploy workflow has been added

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Refactoring / Documentation
- [ ] Version change

if your change is a breaking change, please add `breaking change` label to this PR

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes.

## Does the change impact or break the Docker build?

- [x] Yes
- [ ] No

If Yes: Has Docker been updated accordingly?

- [x] Yes
- [ ] No

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] New and existing unit tests pass locally with my changes


src/cms/data/functions/loadPermissions.js
```diff
@@ -3,11 +3,11 @@
 const createApiUser = async () => {
   const authRole = await findRole("authenticated");
   const password = await strapi.admin.services.auth.hashPassword(
-    process.env.API_USER_PASSWORD
+    process.env.STRAPI_API_USER_PASSWORD
```
marklise, line 6: We don't necessarily have to address in this PR, but I'd like to suggest using `process.env.API_USER_PASSWORD || 'api'` for this line and a similar method on several lines below where we read from env vars but do not have a default (the build will fail without them being set properly).  Perhaps move everything up to the top of the file and set global var there and reference down below so there's a central spot for administrating the env var intake.  My main reason is for local development we can use `sqlite` as a driver instead of using a postgres db, and if we set defaults, the bootstrapping will work out of the box.  The following is a diff of `database.js` that devs can use as a local alternative option (Not sure if we want to add this as a .template or .alternative file in the same dir):

```
diff --git src/cms/config/database.js src/cms/config/database.js
index 864fb3a..4d278df 100644
--- src/cms/config/database.js
+++ src/cms/config/database.js
@@ -2,25 +2,13 @@ module.exports = ({ env }) => ({
   defaultConnection: "default",
   connections: {
     default: {
-      connector: "bookshelf",
+      connector: 'bookshelf',
       settings: {
-        client: "postgres",
-        host: env("DATABASE_HOST"),
-        port: env.int("DATABASE_PORT"),
-        database: env("DATABASE_NAME"),
-        username: env("DATABASE_USERNAME"),
-        password: env("DATABASE_PASSWORD"),
-        ssl: env.bool("DATABASE_SSL"),
-        timezone: "UTC",
+        client: 'sqlite',
+        filename: env('DATABASE_FILENAME', '.tmp/data.db'),
       },
       options: {
-        pool: {
-          min: 0,
-          max: 95,
-          idleTimeoutMillis: 30000,
-          createTimeoutMillis: 30000,
-          acquireTimeoutMillis: 30000,
-        },
+        useNullAsDefault: true,
       },
     },
   },
```

marklise (approved):

nileshnad (changes requested):
The limited CORS restrictions in middleware.js were put in place to restrict use of direct access to these internal CMS URLs. 

Public use of the CMS API is through DataBC endpoints at https://bcparks.api.gov.bc.ca where the CORS restrictions have been removed.

BcGovNeal:
> The limited CORS restrictions in middleware.js were put in place to restrict use of direct access to these internal CMS URLs.
> 
> Public use of the CMS API is through DataBC endpoints at https://bcparks.api.gov.bc.ca where the CORS restrictions have been removed.

Thanks for the clarification Nilesh!  My changes have been restored in https://github.com/bcgov/bcparks.ca/pull/163/commits/3aeea19aa2ccda57005bbe1a43aaeb035d8959eb.  I also made the origins more generic to our namespace

src/cms/data/functions/loadPermissions.js
BcGovNeal: I added a `sqlite` connection and it's the default setting.  For devs that want to use postgres, you can set the `DATABASE_CLIENT=postgres` env

I've also added the API/Admin env defaults in code

https://github.com/bcgov/bcparks.ca/pull/163/commits/f0d7521944aa045c479fb3fbe4161d75444f5335

marklise (approved):

nileshnad (approved):

PR #218, opened 2021-12-17
NOBUG: remove global style overrides

cole:
May break a few things but we can't have everything with
`display: block;`.

### DIFF
--- a/src/staging/src/styles/global.scss
+++ b/src/staging/src/styles/global.scss
@@ -10,13 +10,6 @@ html {
   background-color: white;
   font-family: $globalFont !important;
 }
-* {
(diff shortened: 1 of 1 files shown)


cameronpettit (approved):

DayKirby-OXD (commented):
I'm a bit worried about removing everything in here. E.g. does make sense to remove borders for all elements probably. Maybe just lose the display:block; line (line 18)?

cole:
@DayKirby-OXD The needed resets are provided by bootstrap already, this is just doubling up

DayKirby-OXD (approved):

PR #224, opened 2022-01-11
BRS-181: Make advisories live and filterable

DayKirby-OXD:
# Description

This PR includes the following proposed change(s):

Addresses 3 tix:
- BRS-181 - park advisories live
- BRS-181 - advisories filters
- BRS-262 - keyword search for advisories

Switched from loading all advisories on page load via graphql to fetching only relevant advisories asynchronously
Updated build-public.yaml to point at kong endpoints for api
Added paging so that only a max of 10 advisories are fetched at any given time
Added keyword search, will search all of title, description, park names and event types (all by default, or only selected if any checkboxes checked)
Strapi query to fetch only relevant advisories is built dynamically based on filters
Search is exact string (other than case), including spaces, if any are present (added tix to address in future)
Added flood/wildfire/all dropdown as shortcut to common filter and permalink filters from homepage (changing dropdown reloads page, added ticket to address in future)
Reloading removes all filters (except event type), added ticket to add filter params to url dynamically, which would also create permalinks
Refactored huge alerts code page into several components, put in an advisories folder - page responsible for loading advisories, AdvisoriesList, AdvisoriesCard created for displaying / stying advisories, AdvisoriesFilter created for filtering UI
Updated design, including mobile
Implemented accessibility (other than for header/footer etc, to be addressed elsewhere)


src/staging/src/pages/alerts.js
```diff
@@ -1,133 +1,385 @@
...
+
+    return (q);
+  }, [isKeywordFilter, isParksFilter, isTypesFilter, searchText]);
+
+  const getAdvisories = useCallback((q) => { // q = api query
+
+    let newApiCall = apiUrl + `/public-advisories` + q;
+
+    newApiCall += "&_limit=" + pageLen; // use -1 for unlimited
+    newApiCall += "&_start=" + (pageLen * (pageIndex - 1));
```
cole, line 296: These can be passed to axios as GET params instead of adding them manually, e.g. `.get(newApiCall,  { params: { _limit: pageLen, ...etc } })`

src/staging/src/pages/alerts.js
```diff
@@ -1,133 +1,385 @@
...
+        q += "[eventType.eventType_contains]=" + searchText;
+        n++;
+      }
+      if (useKeywordFilter) {
+        q += "&_where[_or][" + n + "]"
+        q += "[description_contains]=" + searchText;
+        n++;
+        q += "&_where[_or][" + n + "]"
+        q += "[title_contains]=" + searchText;
+      }
```
cole, line 285: alternate style:

```javascript
const params = {};
const orFilters = [];

if (useParksFilter) {
    orFilters.push("[protectedAreas.protectedAreaName_contains]");
}

// ...useTypesFilter, etc

orFilters.forEach((filterName, index) => {
    params[`_where[_or][${index}][${filterName}]`] = searchText;
});
```
Then pass the params to axios.

src/staging/src/components/advisories/advisoryFilter.js
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  const handleTypeFilterChange = (str) => {
+    // This changes the URL query str and causes the page to
+    // rerender with the type changed
+    let newLoc = "/alerts/?type=" + str;
+    if (str !== "public") {
+      newLoc += "s"; // add "s" to wildfire and flood
+    }
+    window.location = newLoc; // this will trigger a reload
```
cole, line 125: I can see we want to update the URL with the new params but the reload feels pretty odd to me. We probably want to add [Reach router](https://www.gatsbyjs.com/docs/reach-router-and-gatsby/) or something so we can have cool URLs everywhere without reloads.

(future enhancement, not for this PR)

src/staging/src/pages/alerts.js
```diff
@@ -138,138 +390,115 @@ const PublicAdvisoryPage = ({ data }) => {
...
-        advisoryNumber
-        dcTicketNumber
-        effectiveDate(formatString: "MMMM DD, YYYY")
-        endDate(formatString: "MMMM DD, YYYY")
-        expiryDate(formatString: "MMMM DD, YYYY")
+/*
+
+  - This section of the graphql query was removed as advisories
+  - are now being loaded through the api
+  - Leaving this here as a comment for future reference
```
cole, line 457: I appreciate the description but still think that this comment is better removed. We have access to old code via git, and what it used to do isn't really relevant anymore.

src/staging/src/components/advisories/advisoryCard.js
```diff
@@ -0,0 +1,258 @@
...
+const AdvisoryCard = ({ advisory, index }) => {
+
+    const classes = useStyles();
+
+    return (
+        <>
+          <Grid item xs={12}
+            className={classes[advisory.detailsClass]}
+            key={advisory.id} thing={advisory.isFirstInYear.toString()}>
+            <div className={((advisory.isFirstInYear || (index === 0)) && (!advisory.advisoryDateObj.dateUnknown)) ? classes.yearHeader : "hidden"}>
```
cole, line 173: I would make this a conditional render instead of using `display: none;` I think if you do hidden, react has to go to the work of putting it into the DOM just to not show it.

cole (approved):
This looks good overall, it's a lot of work! 👏  Few little feedback points but mostly minor and can be deferred given we're at the end of the sprint.

One general thing for next time (probably also applies to my search PR 😐 ) is this is a big chunk of code to review at once, would be easier to have one PR for pulling live data, followed by one for search and filters.

src/staging/src/components/advisories/advisoryPageNav.js
```diff
@@ -0,0 +1,144 @@
...
+    color: "#fff",
+    borderRadius: "5px"
+  },
+  navDots: {
+    fontSize: "1rem",
+    lineHeight: "1rem"
+  },
+}));
+    
+const AdvisoryPageNav = ({ pageIndex, pageCount, setPage }) => {
```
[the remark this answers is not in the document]
DayKirby-OXD: Good catch. Pushing a 1-line fix.

src/staging/src/pages/alerts.js
DayKirby-OXD: for sure. if we remember, let's do this for next related ticket

src/staging/src/components/advisories/advisoryFilter.js
DayKirby-OXD: totally agree a bit clunky, did the simplest thing, knowing that we want to put more params in url in future ticket

cole (approved):

src/staging/src/components/advisories/advisoryList.js
```diff
@@ -0,0 +1,138 @@
...
+    if (capColor === "Grey") {
+      advisory.alertMsg = "Grey Alert - Informational Advisory"
+    } else {
+      advisory.alertMsg = capColor + " Alert - " + level + " Urgency"; // eg Red alert - high urgency
+    }
+
+
+  }
+
+  const processAdvisories = (a) => {
```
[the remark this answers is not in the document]
DayKirby-OXD: k, thanks for the info

PR #227, opened 2022-01-12
BRS 181 - Fix to yaml indentation

DayKirby-OXD:
# Description

- Small fix to ensure advisory count only fetches once per page load
- Small styling fix


cole (approved):

BcGovNeal (changes requested):
The indentations for `.github/workflows/build-public.yaml` are incorrect and is breaking the Github Action workflows.

Please revert the changes and just update the `REACT_APP_CMS_BASE_URL` values without changing the indentations

BcGovNeal (approved):

PR #238, opened 2022-01-20
BRS-351 Update header and footer

DayKirby-OXD:
### Jira Ticket:
BRS-123

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/BRS-123

### Description:
Updated the header and footer
Currently the footer is a large chunk of html being pulled in from strapi
For a few reasons, including the search box that is part of the footer, putting the footer in code
Set up a simple data structure to drive the footer menus, easily replaced with strapi data if we want to make this editable
Search box doesn't function yet, beyond scope of ticket
Small edits to header menu require edits in other environments (detailed in ticket)


DayKirby-OXD:
Need to make a small update to pass jest test. Will update soon

DayKirby-OXD:
Updated jest test, test passed, ready for review

BcGovNeal:
Is it no long necessary to have the header and footer layout be stored in the CMS?

Any particular reason to change the `max-width` on some styles?  The header and footer width aren't the same and they don't match rest of the content width

DayKirby-OXD:
Yes, as discussed in standup, the footer has been moved into code, for a few reasons but especially because the search box is part of the footer and we'll need that in code. However I've set it up so it would be easy to hook the footer menus up to a strapi menu structure similar to the header.

I changed the widths to match the Figma design / Harry's direction. The page contents should also be adjusted via the relevant tickets.

BcGovNeal:
Got it. 

I don't think the footer's width is set correctly if the goal is to have header and footer contents be aligned.  Please see screenshot below.

image

DayKirby-OXD:
Just double checked. Looks like I have the header item spacing wrong. I'll double check the rest while I'm at it, don't approve this yet.

src/staging/src/components/header.js
```diff
@@ -1,13 +1,71 @@
...
+    "& a": { color: "#036" }
+  },
+  infoIcon: {
+    height: "20px;",
+    marginRight: "4px"
+  }
+}));
 
 export default function Header({ children, mode = 'external', content = [] }) {
+  const isBeta = true; // TODO set to false to remove beta banner
```
cole, line 46: Doesn't look like this is checked

src/staging/src/components/header.js
```diff
@@ -1,13 +1,71 @@
...
+          <Box className={classes.linkDivider}>|</Box>
+          <Box className={classes.link}>
+            <a href="#">Help us improve this by submitting here</a>
+          </Box>
+        </Box>
+        <MegaMenu content={content} />
+        </>
     )
   }
+  // TODO is the following case (i.e. mode="external") ever hit?
```
cole, line 67: It looks like it is used on the parks index page (doesn't render for me anyways). I think we can remove.

cole (approved):
Looks good, a few minor nits, and the width issue pointed out above.

DayKirby-OXD:
Made Cole's suggested updates plus a few more minor layout touch-ups.

src/staging/src/components/header.js
DayKirby-OXD: Removed

src/staging/src/components/footer.js
```diff
@@ -1,7 +1,323 @@
...
+  // TODO replace with data from Strapi
+  const menu1 = [
+    { type: "header", display: "Get a permit" },
+    { type: "link", display: "Commercial services", link: "/permits" },
+  ]
+  const menu2 = [
+    { type: "header", display: "Get involved" },
+    { type: "link", display: "Volunteer", link: "/volunteers" },
+    { type: "link", display: "Support BC Parks", link: "/volunteers",
+      note: "TODO update this link",
```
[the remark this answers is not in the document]
DayKirby-OXD: Removed. Covered by another ticket.

src/staging/src/components/header.js
DayKirby-OXD: Removed. I expect the beta banner to be deleted when the time comes, no need to keep it in code with a flag.

BcGovNeal (approved):

PR #240, opened 2022-01-20
BRS-368: fix search sort order

cole:
### Jira Ticket:
BRS-368

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/BRS-368

### Description:
* fix search sort order
* use id from strapi instead of stripping text out of generated id

### DIFF
--- a/src/staging/src/pages/explore.js
+++ b/src/staging/src/pages/explore.js
@@ -168,11 +168,7 @@ export default function Explore({ location, data }) {
     { value: "protectedAreaName:desc", label: "Sort Z-A" },
   ]
 
-  const [sortOption, setSortOption] = useState(
-    location.state && location.state.sortOption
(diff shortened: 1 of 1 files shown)


DayKirby-OXD:
Is ordering by strapi id same as "relevance"?

cole:
The strapiId change is just a cleanup - in Gatsby the data pulled via graphql has an id including the content type (e.g. `Facility-types_17`). It also has a `strapiId` which is the original id from Strapi. I couldn't find the original id before, so I was stripping the text from the generated one. Since the ids are used  to match up, it seems much  better to just use the original.

Not related to search ordering, which is always by name or rank, sorry for the confusion 🙂

DayKirby-OXD (approved):

BcGovNeal (approved):

PR #276, opened 2022-02-15
BRS-422-2 proper image referencing with Gatsby

cameronpettit:
### Jira Ticket:
BRS-422
### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/BRS-422

### Description:
This change makes the 404 page reference link thumbnails via gatsby & graphql queries (the right-er way) rather than referencing an external URL.


marklise (approved):

cole (approved):
Looks good, I like the fade in.

I initially thought that this is what `StaticImage` is for but it looks like you can't pass a prop like `imageLink` as the source value there.

BcGovNeal (approved):

PR #290, opened 2022-02-18
BRS-411 Add ability to mount snowplow from configmap

BcGovNeal:
### Jira Ticket:
BRS-411 

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/BRS-411 

### Description:
I'm using the `react-helmet` package to load the Snowplow script into the `Footer` component.  It will attempt to load it from `/_scripts/snowplow.js` path.  The snowplow.js file is mounted into the container at runtime from OpenShift ConfigMap.  This allows us to use different script per environment and we are able to update them on the fly without rebuilding.

I have created the ConfigMap in dev, test and prod.  So this PR will just work once deployed.  Note that the `snowplow.js` value in prod is currently a placeholder and won't do anything when mounted.

### DIFF
--- a/.github/workflows/build-public.yaml
+++ b/.github/workflows/build-public.yaml
@@ -13,6 +13,7 @@ env:
   IMAGE_NAME: public
   BUILDER_IMAGE_NAME: public-builder
   DEPLOYMENT_NAME: public
+  GATSBY_ENABLE_SNOWPLOW: true
 
 jobs:
   build:
--- a/infrastructure/helm/bcparks/templates/public/public-deployment.yaml
+++ b/infrastructure/helm/bcparks/templates/public/public-deployment.yaml
@@ -39,18 +39,27 @@ spec:
               path: /
               port: 3000
               scheme: HTTP
-            initialDelaySeconds: 10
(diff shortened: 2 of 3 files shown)


cole (changes requested):
I get `expected expression, got '<'` in dev with this branch (snowplow.js returns an HTML 404 page)

Could probably be solved by adding an empty script in the repo that gets mounted over?

The pattern I've used before is to define some env vars at build time (`GATSBY_SNOWPLOW_APP_ID` and `GATSBY_SNOWPLOW_COLLECTOR` or something)  and conditionally include if they are set. This approach looks like it would achieve the same thing though.

BcGovNeal:
Hmm weird, I'm not getting that.  But I am noticing something else in the console that seems to be caused by the missing script.  I'll make some modifications

BcGovNeal:
I added a new `GATSBY_ENABLE_SNOWPLOW` env flag so it'll only attempt to load snowplow if the value is set to `true`

cole (approved):
👍 that works

marklise (approved):

PR #296, opened 2022-02-23
BRS-410: filter out published alerts

cole:
### Jira Ticket:
BRS-410

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/BRS-410

### Description:
Only show advisories for published parks on the alerts page. There was an existing strapi filter, so API behaviour hasn't changed.

### DIFF
--- a/src/staging/src/components/advisories/advisoryCard.js
+++ b/src/staging/src/components/advisories/advisoryCard.js
@@ -208,7 +208,7 @@ const AdvisoryCard = ({ advisory, index }) => {
                     )}
                     <div>
                     {advisory.protectedAreas.length > 0 &&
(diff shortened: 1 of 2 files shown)


marklise (approved):

DayKirby-OXD:
Looks good, just one question. Does the api call that 3rd parties use show all protected areas including the "unpublished" ones (point 2 in the ticket AC)? Maybe add an example of an api call demo'ing this to the ticket?

cole:
@DayKirby-OXD oh, that's a very good point, thanks. I'll add a note explaining that.

DayKirby-OXD (approved):

PR #301, opened 2022-02-25
BRS-410: add isDisplayed flag to protected area

cole:
### Jira Ticket:
BRS-410

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/BRS-410

### Description:
Per further discussion on the ticket, including preview protected areas in the API response via query string was determined to not be sufficient. This PR adds a flag that must be true for the Gatsby site to show the park/protected area. The plan is set this flag whenever a park is ready for display, and then set all (or nearly all) parks to published, so that they are available in the default Strapi API response.

The following queries can be used to do that migration:

```sql
UPDATE protected_areas SET "isDisplayed" = TRUE WHERE published_at IS NOT NULL;
UPDATE protected_areas SET published_at = NOW() WHERE published_at IS NULL;
```

### DIFF
--- a/src/cms/api/protected-area/models/protected-area.settings.json
+++ b/src/cms/api/protected-area/models/protected-area.settings.json
@@ -155,6 +155,11 @@
     },
     "slug": {
       "type": "string"
+    },
+    "isDisplayed": {
+      "type": "boolean",
+      "default": false,
+      "required": true
     }
   }
 }
--- a/src/cms/api/protected-area/services/protected-area.js
+++ b/src/cms/api/protected-area/services/protected-area.js
@@ -147,8 +147,9 @@ module.exports = {
         )
         .groupBy("protected_areas.id");
 
-      // Only include published parks
+      // Only include published & displayed parks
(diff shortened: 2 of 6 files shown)


DayKirby-OXD:
Can't seem to run the branch locally, getting:
```
/Users/daykirby/VSCode/bcparks.ca/src/staging/src/pages/parks.js
  51:37  error  Field "isDisplayed" is not defined by type "StrapiProtectedAreaFilterInput"
```
Have updated the data and refreshed for garibaldi and strathcona, but have not run the above queries (will need more detail if needed)

DayKirby-OXD:
Restarting cms fixed my issue above. Tried some use cases, looks good, will approve.

DayKirby-OXD (approved):

cole:
Now that the beta site live, deploying this change will cause some problems as any parks without `isDisplayed` will disappear, until we have a chance to update the DB. I'm going to split this PR into two, in an attempt to solve this problem.

PR #313, opened 2022-03-09
BRS-339 Match explore page park status with highest precedence advisory

cameronpettit:
### Jira Ticket:
BRS-339

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/BRS-339

### Description:
This ticket changes the accessStatus component on the parks exploration page to match the actual accessStatus of the advisories associated with the park, and not infer accessStatus indirectly.

When searching for parks, the API already returns the titles of the advisories within the park object, so the API has been slightly modified to return the whole advisory, including the accessStatus.

Unfortunately, the structure of an advisory within a park object as mentioned above differs from that of an advisory directly returned via the API,. The React Component `ParkAccessStatus` which displays the access status on the parks details pages was modified to be able to accept the differing structures, so that this component could be also leveraged on the explore page.

### DIFF
--- a/src/staging/src/components/park/parkAccessStatus.js
+++ b/src/staging/src/components/park/parkAccessStatus.js
@@ -1,4 +1,5 @@
 import React from "react"
+import { graphql, useStaticQuery } from "gatsby"
 import PropTypes from "prop-types";
 import { makeStyles } from "@material-ui/core/styles"
 import { Card, CardHeader, Avatar } from "@material-ui/core"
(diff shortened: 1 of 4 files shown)


cole (commented):
This looks good. The only thing I'm concerned about is the `PRECEDENCES` constant that duplicates data in the DB - I don't expect the access status values in the DB to change but if they do, then display on search will break.

One way to improve that would be to do a [`useStaticQuery`](https://www.gatsbyjs.com/docs/how-to/querying-data/use-static-query/) to pull the access status values in via GraphQL. If you prefer to leave that for now I'm fine with that though - this PR fixes the issue in the ticket as is.

cameronpettit:
> 

Good points. I have added `access-statuses` to the collectionTypes array in `gatsby-config` and the accessStatuses are now paired with actual db entries instead of hard-coded ones. Future proofed!

src/staging/src/components/park/parkAccessStatus.js
```diff
@@ -20,18 +21,65 @@ const ICONS = {
...
+        accessStatuses.push({
+          precedence: advisory.accessStatus.precedence,
+          color: advisory.accessStatus.color,
+          text: advisory.accessStatus.accessStatus,
+        });
+      } else {
+        // advisory is coming from explore page
+        // get accessStatus based on precedence
+        let thisStatus = accessStatusList.filter(status => {
+          return status.node.precedence === advisory.accessStatus;
```
cole, line 68: I think this match up should be done by id as that's what the value of `advisory.accessStatus` is (although `id` and `precedence` happen to be the same in most cases). Gatsby calls the id from Strapi `strapiId` so probably `status.node.strapiId === advisory.accessStatus` (you'll have to add `strapiId` to the query as well). The value of `id` includes some text that Gatsby adds so I don't think that will match.

Also up to you but I think `accessStatusList.find` is a minor improvement on filter here.

cameronpettit: Agree with all this (I wasnt even aware of the `accessStatus` vs `strapiId` business). Updated in `dbcc220`

src/staging/src/components/park/parkAccessStatus.js
```diff
@@ -64,8 +65,8 @@ export default function ParkAccessStatus({ advisories }) {
       } else {
         // advisory is coming from explore page
         // get accessStatus based on precedence
-        let thisStatus = accessStatusList.filter(status => {
-          return status.node.precedence === advisory.accessStatus;
+        let thisStatus = accessStatusList.find(status => {
```
cole, line 68: Just need to update `thisStatus[0]` to `thisStatus` below since find returns just the item or null 🙂

cameronpettit: right again!

PR #322, opened 2022-03-14
BRS-459: partial match on park name searches

cole:
### Jira Ticket:
BRS-459

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/BRS-459

### Description:
Add support for partial text matches (e.g. "Rath" for "Rathtrevor")

Also removes `isOpenToPublic` from API results as that has been replaced by displaying the access status text directly.

### DIFF
--- a/src/cms/api/protected-area/services/protected-area.js
+++ b/src/cms/api/protected-area/services/protected-area.js
@@ -151,6 +151,11 @@ module.exports = {
           "protected_areas.orcs",
           "park_operations.orcs"
         )
+        .leftJoin(
+          "park_names",
+          "protected_areas.id",
+          "park_names.protectedArea"
+        )
         .groupBy("protected_areas.id");
 
       // Only include published & displayed parks
@@ -172,7 +177,9 @@ module.exports = {
         // and the description columns of park_activities and park_facilities
         // Any match here counts.
         query.where((builder) => {
-          builder.where(
+          builder.where("protected_areas.protectedAreaName", "ILIKE", `%${searchText}%`);
+          builder.orWhere("park_names.parkName", "ILIKE", `%${searchText}%`);
+          builder.orWhere(
             knex.raw(
               "protected_areas.search_text @@ websearch_to_tsquery('english', ?)",
               [searchText]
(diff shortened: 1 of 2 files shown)


DayKirby-OXD (approved):
Seems to work as expected now. Searched many of the parks that have been frustrating to find, and they now appear.(e.g. "Garibald" or "Portea")

src/cms/api/protected-area/services/protected-area.js
```diff
@@ -166,7 +171,9 @@ module.exports = {
         // and the description columns of park_activities and park_facilities
         // Any match here counts.
         query.where((builder) => {
-          builder.where(
+          builder.where("protected_areas.protectedAreaName", "ILIKE", `%${searchText}%`);
+          builder.orWhere("park_names.parkName", "ILIKE", `%${searchText}%`);
+          builder.orWhere(
```
marklise, line 149: Instead of `LIKE` or `ILIKE`, it would be more better to use indexes to power these queries.  Currently the search is only handling left-handed anchored text and not true FTS /w partials.  `pg_trgm` can assist in this regard when using a GIN in pg.

e.g., `rrow` doesn't return any results.

In addition - getting a HTTP 500 anytime searching locally /w sqlite.  PR #222 is where this was introduced, not this particular commit.  That said, if we're going to abandon `SQLite` then we should spin up a ticket to remove it from our environment and force `pg` on everyone.

cole: Good points, I will look into `pg_trgm` or similar - I'd kind of ruled out using any extensions but it doesn't seem like it would be to complex.

Re: sqlite, that's an unfortunate side effect of using Postgres full text indexes as we do in search. Sqlite doesn't support them, and neither does knex (the db adapter library). I will create a ticket to remove references to sqlite as an option so that all devs are using Postgres.

cole: Updated with GIN indexes. Performance took a hit when making that change, with lots of queries well over 2 seconds, but I played around with it a bit and switched from one big JOIN/GROUP BY to a lot of subqueries and Postgres seems able to handle that better. It makes this change much bigger but I think it's an improvement.

Performance is still not ideal, seeing 1.5s or so for queries with a lot of results when I enable all parks (e.g. a query for "Park"). However most searches seem to be sub 500ms and "rrow" now brings up Arrow Lake Park. Also "garybaldi" brings up Garibaldi, etc.

marklise (approved):
👍

DayKirby-OXD:
Picks up many more mispellings, and seems much faster. Nice work!

PR #323, opened 2022-03-14
BRS-408 pagination reset after filter update on mobile view

cameronpettit:
### Jira Ticket:
BRS-408

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/BRS-408

### Description:
This change resets the park pagination to page 1 after filters are updated. The `searchFilter` component is used in mobile view which has different handler functions for when a filter is checked/unchecked. This change passes the `setCurrentPage` state function from the explore page to the `searchFilter` component so the currentPage state can be tracked there.

### DIFF
--- a/src/staging/src/components/search/searchFilter.js
+++ b/src/staging/src/components/search/searchFilter.js
@@ -115,6 +115,7 @@ const SearchFilter = ({
     selectedFacilities,])
 
   const searchParkFilter = () => {
+    setCurrentPage(1);
     navigate("/explore", {
       state: {
         selectedActivities,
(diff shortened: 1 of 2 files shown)


DayKirby-OXD (changes requested):
Changing filter seems to work as desired, but found a case that's maybe unhandled. In mobile, searched "lake", then went to page 3. Then went to top and searched "mountain", still on page 3

cameronpettit:
> Changing filter seems to work as desired, but found a case that's maybe unhandled. In mobile, searched "lake", then went to page 3. Then went to top and searched "mountain", still on page 3

This issue was also present while using text search in desktop view. Latest commit should have fixed the issue in desktop view, mobile view, and in the text search in the mobile modal.

DayKirby-OXD (approved):
Looks good, thx.

marklise (approved):

PR #325, opened 2022-03-16
BRS-361 update footer

DayKirby-OXD:
### Jira Ticket:
BRS-361

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/BRS-361

### Description:
- Update footer links with correct text and urls (note that some links dead outside of prod)
- Convert MUI to Bootstrap, moved footer styles to own sheet, removed unused rules
- Updated jest screenshot

### DIFF
--- a/src/staging/__tests__/components/__snapshots__/footer.test.js.snap
+++ b/src/staging/__tests__/components/__snapshots__/footer.test.js.snap
@@ -9,128 +9,88 @@ exports[`Footer renders correctly 1`] = `
     id="home-footer"
   >
     <div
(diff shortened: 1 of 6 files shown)


cole (approved):
Looks good, nice work removing MUI

cole:
@cameronpettit re: why is the footer the only component we have Jest scripts for - I think it's kind of an accident, we added the testing framework and my guess is just it seemed like an easy example test. Footer turned out to be not a great choice unfortunately. Ideally we'll add more component tests in future so it will make more sense then 🤞

PR #329, opened 2022-03-17
BRS-457 review of incorrectly cascading global styles

cameronpettit:
### Jira Ticket:
BRS-457

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/BRS-457

### Description:
This ticked was to investigate cases where some global styles were incorrectly cascading down to child elements. Upon review, it appears that the specific issue involving `betaLanding.scss` had already been rectified. This file was removed as its contents are no longer referenced anywhere. 
A quick scan of the other scss files did not immediately reveal similar cases where styles may be cascading incorrectly, but if future styling issues are uncovered, they can be specifically addressed quite easily on a case-by-case basis in future tickets.


cameronpettit:
Incorrect implementation. will open new PR

PR #330, opened 2022-03-18
BRS-457 review of incorrectly cascading intro page styles

cameronpettit:
### Jira Ticket:
BRS-457

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/BRS-457

### Description:
Some scss styles directly applied to html elements at the root level in `betaLanding.scss` were leaking into other parts of the site. This change wraps the contents of `betaLanding.scss` in a class that can be applied only to the intro page, to properly segregate the styles.

A quick scan of the other scss files did not immediately reveal similar cases where styles may be cascading incorrectly, but if future styling issues are uncovered, they can be specifically addressed quite easily on a case-by-case basis in future tickets.

### DIFF
--- a/src/staging/src/pages/intro.js
+++ b/src/staging/src/pages/intro.js
@@ -37,7 +37,7 @@ const IntroPage = ({ data }) => {
         <Header mode="internal" content={menuContent} />
       </div>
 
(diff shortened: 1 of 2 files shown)


marklise (approved):

DayKirby-OXD (approved):
Nit, feel free to ignore. Might be better if intro-page was an id instead of a class, as it's intended as a unique element. Approved.

PR #338, opened 2022-03-24
BRS-473: redirect added to Caddyfile for /map, map on park pages now links to BC Parks Map

meredithom:
### Jira Ticket:
BRS-473

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/BRS-473

### Description:
This is a description.

### DIFF
--- a/src/staging/Caddyfile
+++ b/src/staging/Caddyfile
@@ -6,6 +6,8 @@
 
     root * /usr/share/caddy
 
+    redir /map https://governmentofbc.maps.arcgis.com/apps/webappviewer/index.html?id=077ef73a1eae4ca88f2bafbb831215af&query=British_Columbia_Parks_Ecological_Reserves_and_Protected_Areas_8747,ORCS_PRIMARY,0000
+
     handle_errors {
       @404 {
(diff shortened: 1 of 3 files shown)


cole (commented):
Linking the inline behaves a bit strangely because there are still working controls on the map but clicking on them links you away from the page. Recommend linking the header (or some other text) only, or removing the map control widgets. Code looks fine, just odd behaviour.

Caddy syntax looks correct, we'll have to check it's working on dev.

meredithom:
> Linking the inline behaves a bit strangely because there are still working controls on the map but clicking on them links you away from the page. Recommend linking the header (or some other text) only, or removing the map control widgets. Code looks fine, just odd behaviour.
> 
> Caddy syntax looks correct, we'll have to check it's working on dev.

Agree that it behaved strangely. I've now removed the UI and scroll functionality so the map appears static 
https://github.com/bcgov/bcparks.ca/pull/338/commits/64224c867cc35cb30f0bd769f17952cc741c8b3a

marklise (approved):

danieltruong (approved):

PR #349, opened 2022-03-30
BRS-476 update to operation dates

DayKirby-OXD:
### Jira Ticket:
BRS-476

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/BRS-476

### Description:
- switch from table to accordions, add dates to accordion content
- update dates area to include park status (plus access status refactor)
- link from subareas to facilities (plus add anchors to facilities)
- note that links only work when facilities have codes in cms content (see ticket)


src/staging/src/components/park/parkFacility.js
```diff
@@ -11,61 +12,111 @@ import StaticIcon from "./staticIcon"
...
 
-  if (facilityData.length === 0) return null
+  const toggleExpand = useCallback(
+    index => {
+      expanded[index] = !expanded[index]
+      setExpanded([...expanded])
+    },
+    [expanded]
+  )
+  //const [isInitial, setIsInitial] = useState(true)
```
cole, line 28: Remove commented line or add explanation

src/staging/src/components/park/parkFacility.js
```diff
@@ -11,61 +12,111 @@ import StaticIcon from "./staticIcon"
...
 
-  const toggleExpand = index => {
-    expanded[index] = !expanded[index]
-    setExpanded([...expanded])
-  }
+  useEffect(() => {
+    // Each element in the array needs a unique id for the accordions
+    // to be togglable through code
+    let i = 0
+    for (const idx in facilityData) {
```
cole, line 55: I don't think it causes any problems here but for...in is not ideal for arrays, just FYI: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#array_iteration_and_for...in

probably a `forEach` case or standard for loop

src/staging/src/components/park/parkDates.js
```diff
@@ -66,25 +92,28 @@ export default function ParkDates({ data }) {
...
+            dateRec.offSeasonEndDate
+          )
 
-          dates.push({
-            dateCount: saDateCount,
-            subArea: subArea.parkSubArea,
-            serviceDates: serviceDates,
-            resDates: resDates,
-          })
+          if (saDateCount === 1) {
```
cole, line 100: I'm not sure if there are any cases where it's an issue but I think this logic will result in a leading comma if there are two dates and the first is inactive.

Recommend building 3 arrays of active date strings and using `.join(" ,")`

src/staging/src/components/park/parkDates.js
```diff
@@ -1,18 +1,39 @@
...
+import Container from "react-bootstrap/Container"
+import Heading from "./heading"
+import HtmlContent from "./htmlContent"
+import StaticIcon from "./staticIcon"
+import { ParkAccessFromAdvisories } from "../../components/park/parkAccessStatus"
 
 export default function ParkDates({ data }) {
   const parkOperation = data.parkOperation || {}
   const subAreas = data.subAreas || []
+  subAreas.sort((a, b) => (a.parkSubArea >= b.parkSubArea ? 1 : -1))
```
cole, line 17: I think it's fine for now, just an FYI for future: there's a lot of prop mutation happening in this component. `data.subAreas` gets sorted in place, each subarea gets a few properties added, etc. These changes will propagate up to the parent component. It's not a big issue because this component is only used once on the page, but it violates the principle of read only props:
https://reactjs.org/docs/components-and-props.html#props-are-read-only

src/staging/src/components/park/parkFacility.js
```diff
@@ -11,61 +12,111 @@ import StaticIcon from "./staticIcon"
...
+  )
+  //const [isInitial, setIsInitial] = useState(true)
+
+  const checkHash = useCallback(() => {
+    // Check hash in url
+    // if we find a matching facilityCode, open that facility accordion
+    // TODO setup useEffect etc.
+    let h = ""
+
+    if (typeof window !== "undefined") {
```
cole, line 36: Recommend using useLocation from reach router (used by Gatsby) instead of checking window:
```
import { useLocation } from "@reach/router"

...
const location = useLocation()
const hash = location.hash
```

Should update without the need for a manual event listener

cole (approved):
Approving with comments, mostly nitpicks - leaving any changes to your discretion.

One thing to note is we'll need to make sure `facilityCode` is set on dev before merging.

cole:
FYI for other reviewers, you can set facility codes in postgres via:
`UPDATE facility_types SET "facilityCode" = REPLACE(lower("facilityName"), ' ', '-');`

src/staging/src/components/park/parkDates.js
DayKirby-OXD: Moving the increment into the isActive condition, will avoid the case with first inactive date

src/staging/src/components/park/staticIcon.js
```diff
@@ -9,19 +9,16 @@ export default function StaticIcon({ name, size }) {
   if (name && name.startsWith("https://")) {
     iconSrc = name
   } else if (name) {
-    iconSrc = `/icons/${name}.svg`
+    if (name.indexOf(".svg") === -1) {
```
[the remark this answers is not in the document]
DayKirby-OXD: Done

src/staging/src/components/park/parkFacility.js
DayKirby-OXD: Done

src/staging/src/components/park/parkDates.js
```diff
@@ -25,35 +46,40 @@ export default function ParkDates({ data }) {
...
-
-  let dates = []
 
   for (let idx in subAreas) {
     const subArea = subAreas[idx]
 
     if (subArea.isActive) {
+      const typeObj = subArea.parkSubAreaType || {}
+      const iconUrl = typeObj.iconUrl || ""
+      const typeIcon = iconUrl.split("/")[iconUrl.split("/").length - 1] // ignore path, get filename
```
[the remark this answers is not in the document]
DayKirby-OXD: This approach gave me errors, going to leave as is.

src/staging/src/components/park/parkFacility.js
DayKirby-OXD: Done.

src/staging/src/components/park/parkDates.js
DayKirby-OXD: Updated.

src/staging/src/components/park/parkFacility.js
DayKirby-OXD: I could find a simple way to use this approach without causing a lot of refactoring and/or "too many rerender" errors. Leaving as is.

cole (approved):

marklise (approved):

PR #350, opened 2022-03-30
BRS-434 update correct park type on park details page

mayurmarakana89:
### Jira Ticket:
BRS-434

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/BRS-434

### Description:
Update correct park type on park details page


src/staging/src/components/park/about.js
```diff
@@ -1,5 +1,9 @@
 import React from "react"
+<<<<<<< Updated upstream
 import { parseISO, format } from "date-fns"
+=======
+import { capitalize } from "lodash"
+>>>>>>> Stashed changes
```
cole, line 6: Looks like a merge that didn't get resolved here

cole (changes requested):
Looks good other than the merge error 🚀

src/staging/src/components/park/about.js
mayurmarakana89: Oops! Good catch! Have pushed the fix.

cole (approved):

marklise (commented):
Not a huge fan of importing `_`.  There are simple ways to deal with both `sortBy` and `capitalize` in ES6.  Examples below:

```
const capitalize = ([firstLetter, ...restOfWord]) => firstLetter.toUpperCase() + restOfWord.join("");
```

```
const fruits = [
  {name:"banana", amount: 2},
  {name:"apple", amount: 4},
  {name:"pineapple", amount: 2},
  {name:"mango", amount: 1}
];

const sortBy = (key) => {
  return (a, b) => (a[key] > b[key]) ? 1 : ((b[key] > a[key]) ? -1 : 0);
};

fruits.concat().sort(sortBy("name"));

```

src/staging/src/utils/helpers.js
```diff
@@ -0,0 +1,5 @@
+export const capitalizeFirstLetter = (str) => str.charAt(0).toUpperCase() + str.slice(1).toLowerCase();
+
+export const sortByKey = (key) => {
```
DayKirby-OXD, line 3: Nit: not seeing any reference to sortByKey. Could be removed, optional.

DayKirby-OXD (approved):
Looks good. At some point we might want ParkTypes in the CMS as a content type, but that seems like overkill at the moment, and definitely beyond scope of the ticket.

src/staging/src/utils/helpers.js
mayurmarakana89: Good catch! Removed the unused helper function.

PR #363, opened 2022-04-13
BWT-194 updated parks detail menu

DayKirby-OXD:
### Jira Ticket:
BWT-194

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/BWT-194

### Description:
Addressed mobile menu for park detail not being sticky by:
Updated park detail floating menu to use same component as long static pages
Also fixed a couple of broken references in the menu

### DIFF
--- a/src/staging/src/templates/park.js
+++ b/src/staging/src/templates/park.js
@@ -14,7 +14,7 @@ import {
 import { makeStyles } from "@material-ui/core/styles"
 import useScrollSpy from "react-use-scrollspy"
 
-import { capitalizeFirstLetter } from "../utils/helpers";
+import { capitalizeFirstLetter } from "../utils/helpers"
 
 import Footer from "../components/footer"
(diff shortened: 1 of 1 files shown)


src/staging/src/templates/park.js
```diff
@@ -14,7 +14,7 @@ import {
 import { makeStyles } from "@material-ui/core/styles"
 import useScrollSpy from "react-use-scrollspy"
 
-import { capitalizeFirstLetter } from "../utils/helpers";
+import { capitalizeFirstLetter } from "../utils/helpers"
```
marklise, line 17: Are there files missing from this PR? It doesn't seem to be addressing the bug.

DayKirby-OXD: Oh weird! Yes there should be more than that. Mark I'll reach out if I can't figure it out

PR #379, opened 2022-07-28
BRS-732: BCparks api maxing out and causing slow - unresponsiveness i…

marklise:
…n alerts/advisories

### Jira Ticket:
BRS-732

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/BRS-732

### Description:
This changes the way park-access-statues api call works.  It used to be a large and expensive operation.  Now, it is done in a cron once every 5 minutes, instead of every time a pageload of https://bcparks.ca/wildfire or the https://bcparks.ca/explore/parks page (or others).   The cron saves this cached payload into a new record type in strapi.  There is a new endpoint created to pull the cached data that the cron created.  The old endpoint is now force limited to `_limit: 10` to avoid any future performance issues.  Consumers of the old endpoint should use the new cached endpoint.

New endpoint:
```[2022-07-28T17:14:45.031Z] debug GET /park-access-statuses-cache (52 ms)``` 200

Old endpoint:
```[2022-07-28T17:15:18.373Z] debug GET /park-access-statuses?_limit=1200 (1940 ms) 200```

### DIFF
--- a/src/cms/api/park-access-status-cache/config/routes.json
+++ b/src/cms/api/park-access-status-cache/config/routes.json
@@ -1,0 +1,12 @@
+{
+  "routes": [
+    {
+      "method": "GET",
+      "path": "/park-access-statuses-cache",
+      "handler": "park-access-status-cache.find",
+      "config": {
+        "policies": []
+      }
+    }
+  ]
+}
--- a/src/cms/api/park-access-status-cache/controllers/park-access-status-cache.js
+++ b/src/cms/api/park-access-status-cache/controllers/park-access-status-cache.js
@@ -1,0 +1,8 @@
+"use strict";
+
+module.exports = {
+  async find(ctx) {
+    console.log("Park access status cache find:", ctx);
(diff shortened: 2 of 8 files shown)


danieltruong (approved):

cameronpettit (approved):

mayurmarakana89:
Changes looks good, just one comment on console logs, would it be safe on PROD to print values or these are just for testing purposes?

marklise:
> Changes looks good, just one comment on console logs, would it be safe on PROD to print values or these are just for testing purposes?

I'm wanting to continue to view the logs with detail in PROD just in case we have other cases to deal with.

mayurmarakana89:
> > Changes looks good, just one comment on console logs, would it be safe on PROD to print values or these are just for testing purposes?
> 
> I'm wanting to continue to view the logs with detail in PROD just in case we have other cases to deal with.

Sounds good as long as it’s basically for tracing the issue.

PR #392, opened 2022-09-23
BWT-230: Add missing back-end functionality and decoded entities in UI

molund:
### Jira Ticket:
BRS-230

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/BRS-230

### Description:
- Implemented the missing parts of previous commit that were causing the build to fail
- Show the unescaped entities in the table and in the dropdown
- Added park-names/items API endpoint for lighter-weight retrieval of park names


foley-lynn (approved):

src/admin/package.json
```diff
@@ -24,6 +24,7 @@
     "babel-loader": "8.1.0",
     "babel-plugin-transform-export-extensions": "^6.22.0",
     "bootstrap": "^4.6.0",
+    "he": "^1.2.0",
```
marklise, line 27: This package is very old and doesn't appear to be well maintained.  A quick search found https://github.com/mdevils/html-entities which is a lot more active and may be a better choice.

molund: Thanks @marklise. I swapped it to html-entities and it seems to behave exactly the same as he.

marklise (approved):

PR #403, opened 2022-11-02
CM-00: hotfix to fix tests failing

foley-lynn:
### Jira Ticket:
None

### Jira Ticket URL:
None

### Description:
This is a fix to address Strapi tests failing each time.

### DIFF
--- a/src/cms/config/functions/responses/404.js
+++ b/src/cms/config/functions/responses/404.js
@@ -1,5 +1,5 @@
 "use strict";
 
-module.exports = async (/* ctx */) => {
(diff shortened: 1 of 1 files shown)


foley-lynn:
The test that was running didn't like that this file was commented out. I don't think it actually does anything in our implementation.

We should circle back and see if we can figure out why this happened all of the sudden. I haven't tracked that down, but I think it's more likely an updated dependency than anything we did.

nataliasabadysh (approved):

molund (approved):

PR #407, opened 2022-11-07
Bwt 349: park-access-statuses API endpoint filter for accessStatus field

nataliasabadysh:
Jira Ticket:
Bwt 349

Jira Ticket URL:
https://bcparksdigital.atlassian.net/jira/software/projects/BWT/boards/5?assignee=633db6bb61dbef2805c1f25c&selectedIssue=BWT-349

### DIFF
--- a/src/cms/api/protected-area/custom/protected-area-status.js
+++ b/src/cms/api/protected-area/custom/protected-area-status.js
@@ -62,10 +62,12 @@ const getPublishedPublicAdvisories = async () => {
 // custom route for park status view
 const getProtectedAreaStatus = async (ctx) => {
   let entities;
+  const { accessStatus, ...query } = ctx.query
+
   if (ctx.query._q) {
     entities = await strapi.services["protected-area"].search(ctx.query);
   } else {
-    entities = await strapi.services["protected-area"].find(ctx.query);
+    entities = await strapi.services["protected-area"].find(query);
   }
 
   const regionsData = await strapi.services["region"].find({ _limit: -1 });
@@ -93,7 +95,7 @@ const getProtectedAreaStatus = async (ctx) => {
 
   const publicAdvisories = await getPublishedPublicAdvisories();
 
-  const payload = entities.map((protectedArea) => {
+  let payload = entities.map((protectedArea) => {
     let publicAdvisory = getPublicAdvisory(
       publicAdvisories,
       protectedArea.orcs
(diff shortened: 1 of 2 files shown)


molund:
@nataliasabadysh would it be fairly easy to also add support for a `accessStatus_ne` parameter?
e.g. `?accessStatus_ne=Open`
I think this option might accomplish the original goal of the ticket description better than checking for 9 different values.

1. Open
2. Full closure
3. Closed to the public
4. Inaccessible
5. Park is under evacuation alert
6. Park is inaccessible
7. Restricted; permit required
8. Park is under evacuation order
9. Park is within a fire perimeter
10. Warning

molund (changes requested):
@nataliasabadysh would it be fairly easy to also add support for a `accessStatus_ne` (not equal to) parameter?
e.g. `?accessStatus_ne=Open`
I think this option might accomplish the original goal of the ticket description better than checking for 9 different values.

src/cms/api/protected-area/custom/protected-area-status.js
```diff
@@ -270,6 +272,17 @@ const getProtectedAreaStatus = async (ctx) => {
...
+      (o) => o?.accessStatus?.toLowerCase() == accessStatus.toLowerCase()
+    );
+  } else if (accessStatus_ne) {
+    return payload = payload.filter(
+      (o) => o?.accessStatus?.toLowerCase() != accessStatus_ne.toLowerCase()
+    );
+  }
+
   // Store unfiltered payload into a cached location (unfiltered == no orcs specified)
   if (!ctx.query.orcs) {
```
molund, line 287: I think we need to update this to include accessStatus and accessStatus_ne (in additon to orcs).

molund (changes requested):

src/cms/api/protected-area/custom/protected-area-status.js
nataliasabadysh: yes, I added a case with the both 
please have a look 
thank you

ayumi-oxd (approved):
LGTM, if you modify the condition part in the filter that @molund mentioned in the meeting.

molund (approved):

molund (changes requested):
This if condition should be updated so the ParkAccessStatusCaches collection only gets updated when there is an unfiltered query.....
` if (!ctx.query.orcs) {`

There is a cron job that runs every 5 minutes to update ParkAccessStatusCaches

molund (approved):

PR #408, opened 2022-11-10
CM-30: added title field to page

ayumi-oxd:
### Jira Ticket:
CM-30

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/CM-30

### Description:
Added a new field called Title at the Pages.

### DIFF
--- a/src/cms/api/page/models/page.settings.json
+++ b/src/cms/api/page/models/page.settings.json
@@ -14,6 +14,10 @@
     "Slug": {
       "type": "string"
     },
+    "Title": {
+      "type": "string",
+      "required": false
+    },
     "Template": {
       "type": "enumeration",
       "enum": [
--- a/src/staging/gatsby-node.js
+++ b/src/staging/gatsby-node.js
@@ -226,6 +226,7 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
         Slug
         Template
         Content
+        Title
       }
     }
   }
--- a/src/staging/src/templates/staticLanding1.js
+++ b/src/staging/src/templates/staticLanding1.js
@@ -1,7 +1,7 @@
 import React from "react"
-import { useStaticQuery, graphql } from "gatsby"
+import { useStaticQuery, graphql, Link } from "gatsby"
 import { StaticImage } from "gatsby-plugin-image"
-import { CssBaseline } from "@material-ui/core"
+import { CssBaseline, Breadcrumbs } from "@material-ui/core"
 
 import Footer from "../components/footer"
 import Header from "../components/header"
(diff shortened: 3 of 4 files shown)


src/staging/gatsby-node.js
```diff
@@ -226,6 +226,7 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
         Slug
         Template
         Content
+        Title
```
ayumi-oxd, line 229: Added `Title` here to make it available on FE.

src/staging/src/templates/staticLanding1.js
```diff
@@ -101,6 +101,21 @@ const LandingPage = ({ pageContext }) => {
...
+              {renderBreadcrumbs(menuContent, pageContext?.page)}
+            </Breadcrumbs>
+          </div>
+          <div className="static-content--header">
+            <h1 className="header-title">
+              {page.Title}
+            </h1>
+          </div>
+        </>
+      )}
```
ayumi-oxd, line 118: Added breadcrumbs and title even if the page does not have `HTMLArea`

src/staging/src/templates/staticLanding1.js
```diff
@@ -121,4 +136,31 @@ const LandingPage = ({ pageContext }) => {
...
+          {parent.title}
+        </Link>
+      )
+      parent = menuContent.find(mc => mc.strapiId === parent?.strapiParent?.id)
+      return addItems(parent, menuContent, breadcrumbItems)
+    }
+    return breadcrumbItems.reverse()
+  }
+}
+
```
ayumi-oxd, line 165: Added this to render breadcrumbs from `Menus`.

src/cms/api/page/models/page.settings.json
```diff
@@ -33,6 +33,10 @@
         "parks.card-set",
         "parks.fancy-link"
       ]
+    },
+    "Title": {
+      "type": "string",
+      "required": true
```
foley-lynn, line 39: Let's make this not required, please.

src/staging/src/templates/staticLanding1.js
foley-lynn: Hmmm... This won't ever really be used, but it's a good reminder of how easy it will be to add the header via component. Nice.

foley-lynn:
This seems to work well. That said, it may require some manual work on each environment to get it working. I'll out line those steps in the ticket as well, and Ayumi you can confirm them.

Steps to get it running locally:
Once you've checked out the branch, in your backend, run `yarn start` instead of the usual `yarn develop`. This is required to register the field in Strapi.
Next step is to go into the Strapi admin, open a page, and then add something to the new Title field (likely at the bottom).
Final step is to Start Gatsby as you normally would. Then you can see on Static Content pages that if you add a Title, it'll override the PageTitle.

src/staging/src/templates/staticLanding1.js
ayumi-oxd: Yes, it's for the very edge case that the `staticLanding` page doesn't have any elements. This part will be changed once [this ticket](https://bcparksdigital.atlassian.net/browse/CM-143) is implemented 🙂

molund (approved):

PR #426, opened 2022-11-29
CM-9: Parks without lat/long

molund:
### Jira Ticket:
CM-9

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/CM-9

### Description:
- Hide the map if a park doesn't have a latitude and longitude
- Hide the entire location section if the park doesn't have either a latitude and longitude or a location note


ayumi-oxd (approved):
LGTM 👍 Checked the Maps and Location section was hidden when there was no lat/long data.

nataliasabadysh (approved):

PR #432, opened 2022-12-05
CM-110: changed certain protectedArea fields from text to richtext

foley-lynn:
### Jira Ticket:
CM-110

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/CM-110

### Description:
Certain fields in the protectedArea collection type should be converted from plain text to rich text.

### DIFF
--- a/src/cms/api/park-activity/models/park-activity.settings.json
+++ b/src/cms/api/park-activity/models/park-activity.settings.json
@@ -15,7 +15,7 @@
       "type": "string"
     },
     "description": {
-      "type": "text"
+      "type": "richtext"
     },
     "isActivityOpen": {
       "type": "boolean"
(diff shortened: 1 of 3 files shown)


molund:
I tested it, and it seems to work, but I'm wondering if we should update a couple of other fields at the same time because we won't have a lot of time to do multiple rounds of QA and UAT before the data import.  

We will be importing tens of thousands of records into these 2 fields, so they are probably the highest priority....
- ParkActivity.description
- ParkFacility.description

ayumi-oxd (approved):

molund (approved):

foley-lynn:
@molund updated as requested. Can you take a quick peek to make sure I got it right?

PR #437, opened 2022-12-07
NOBUG: Refactor config file

nataliasabadysh:
moved back a separate s3 config  file for production


src/cms/config/env/test/production/plugins.js
```diff
@@ -0,0 +1,14 @@
+module.exports = ({ env }) => ({
+  upload: {
```
molund, line 2: Can you please include the breakpoints as well?
```
    breakpoints: {
      small: 720,
    },

molund (changes requested):

molund (approved):

foley-lynn (approved):

PR #451, opened 2022-12-12
CM-162: Add site template

ayumi-oxd:
### Jira Ticket:
CM-162

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/CM-162

### Description:
- Added new fields and a new relation: `slug`, `hasDayUsePass`, `description`, `locationNotes`, and `parkOperation`
- Added a new path for site page: `/protectedArea.slug/site.siteName` eg. `/parks/mount-robson/berg-lake-trail`
- Added a site template based on the design mockup.
- Those new fields  `slug`, `hasDayUsePass`, `description`, and `locationNotes` are available on Strapi, but won't be available on Gatsby (front-end) at this moment to avoid the build error.

### DIFF
--- a/src/staging/gatsby-node.js
+++ b/src/staging/gatsby-node.js
@@ -294,8 +294,10 @@ async function createSites({ graphql, actions, reporter }) {
   const result = await strapiApiRequest(graphql, siteQuery)
 
   result.data.allStrapiSites.nodes.forEach(site => {
+    // TODO: slug can be deleted when site.slug has created on all strapi env
     const slug = slugify(site.siteName).toLowerCase()
     const parkPath = site.protectedArea?.urlPath
+    // TODO: change sitePath `${parkPath}/${slug}` to `${parkPath}/${site.slug}` when site.slug has created on all strapi env
     const sitePath = `${parkPath}/${slug}`
     actions.createPage({
       path: sitePath,
--- a/src/staging/gatsby-node.js
+++ b/src/staging/gatsby-node.js
@@ -213,6 +213,16 @@ exports.createSchemaCustomization = ({ actions }) => {
     url: String
     imgUrl: String
   }
+
+  type StrapiSites implements Node {
+    siteName: String
+    siteNumber: Int
+    orcsSiteNumber: String
+    protectedArea: StrapiProtectedArea
+    parkActivities: [StrapiParkActivities]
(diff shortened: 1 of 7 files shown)


src/staging/gatsby-node.js
```diff
@@ -263,6 +274,37 @@ async function createParks({ graphql, actions, reporter }) {
...
+        }
+      }
+      totalCount
+    }
+  }
+  `
+  const result = await strapiApiRequest(graphql, siteQuery)
+
+  result.data.allStrapiSites.nodes.forEach(site => {
+    const slug = slugify(site.siteName).toLowerCase()
```
foley-lynn, line 297: Do we follow this pattern elsewhere? I don't like it, it feels like generally on this site we set the slug manually, and I'd like to continue with that patten here for consistency.

So instead of using the page title, I would prefer to add a `siteUrl` field that could be set manually.

If the issue is that we can't grab that with an import, perhaps we could add the new field, and use the current scheme as a fallback.

Feel free to discuss offline if there's a reason I can't see.

src/staging/src/templates/site.js
```diff
@@ -0,0 +1,500 @@
...
+import ParkFacility from "../components/park/parkFacility"
+import ParkHeader from "../components/park/parkHeader"
+import ParkOverview from "../components/park/parkOverview"
+import ParkPhotoGallery from "../components/park/parkPhotoGallery"
+import MapLocation from "../components/park/mapLocation"
+import ScrollToTop from "../components/scrollToTop"
+
+import "../styles/parks.scss"
+
+const drawerWidth = 230
```
foley-lynn, line 34: Since this variable is no longer scoped to a single template, it should probably be abstracted further back. Not a big deal since we'll be re-writing this template in the not-too-distant future, but good practice.

foley-lynn (approved):
Approved once comments are addressed.

src/staging/gatsby-node.js
ayumi-oxd: This is because a new field won't be available on Gatsby when this PR will be merged.
To avoid the build error, I can create and include a new field `siteUrl` or `slug` in this commit but applying the new field to this site path logic will be after this ticket deployment.

molund (approved):

PR #457, opened 2022-12-14
[WIP]: Redirect and 404 handling

nataliasabadysh:
### Jira Ticket:
CM-134

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/jira/software/projects/CM/boards/11?assignee=633db6bb61dbef2805c1f25c&selectedIssue=CM-134


src/cms/data/functions/seedData.js
```diff
@@ -28,6 +28,7 @@ const loadData = async () => {
       otherData.loadUrgency(),
       otherData.loadFireCentre(),
       otherData.loadFireZone(),
+      otherData.loadRedirects(),
```
molund, line 31: If we are going to load redirects as part of the data import, then they should also be deleted and loaded in **reimportData.js**.  

Add these lines where you see `park-name` being deleted and reloaded:
`otherData.loadRedirects(),`
`strapi.services["redirects"].delete(),`

src/cms/api/redirects/models/redirects.settings.json
```diff
@@ -0,0 +1,20 @@
+{
+  "kind": "collectionType",
+  "collectionName": "redirects",
```
molund, line 3: `redirects`  shouldn't be plural.  All the other collection types names are singular.

We might even want to call it something like `legacy-redirect` instead, since **redirect** is almost like a reserved word in web development.

src/cms/api/redirects/models/redirects.settings.json
```diff
@@ -0,0 +1,20 @@
...
+  "info": {
+    "name": "Redirects"
```
molund, line 5: If you change the name to legacy-redirect above, then this should be changed to LegacyRedirects

molund:
Can you please add at least one sample record to src/cms/data/redirects.json

I would add a comment there, but I can't comment on an empty file in github

src/staging/package.json
```diff
@@ -20,11 +20,13 @@
     "date-fns": "^2.28.0",
     "gatsby": "^3.13.1",
     "gatsby-image": "^3.11.0",
+    "gatsby-plugin-client-side-redirect": "^1.1.0",
```
molund, line 23: Can you confirm that this is being used?

src/staging/gatsby-node.js
```diff
@@ -231,10 +237,65 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
...
+  let parks = resultPark.data.allStrapiProtectedArea.nodes;
+
+  parks.forEach((park) => {
+    redirects.map((item) => {
+
+      if (park.slug !== item.toPath) {
+        return actions.createRedirect({
+          fromPath: item.fromPath,
+          toPath: item.toPath,
+          isPermanent: true,
```
molund, line 287: We really just need fromPath and toPath.  The other options are for server-side redirects.

PR #468, opened 2022-12-16
CM-134: Redirect FE

nataliasabadysh:
### Jira Ticket URL:
https://bcparksdigital.atlassian.net/jira/software/projects/CM/boards/11?assignee=633db6bb61dbef2805c1f25c&selectedIssue=CM-134


src/staging/gatsby-node.js
```diff
@@ -254,6 +259,52 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
...
+
+  parks.map(park => {
+    redirects.map(item => {
+      if (park.slug !== item.toPath) {
+        return actions.createRedirect({
+          fromPath: item.fromPath,
+          toPath: item.toPath,
+        })
+      }
+    })
```
molund, line 306: This should just be two loops instead of a nested loop

1.  Loop through all the redirects and create a redirect using fromPath and toPath

2. Loop through all the parks and create a redirect for each using "the path from the oldUrl" as the fromPath and the slug as the toPath   (but only if they are different)

e,g, 
- **oldUrl** = https://bcparks.ca/eco_reserve/ambrose_er.html
- **slug** = ecological-reserves/ambrose-lake

therefore
- **fromPath** = /eco_reserve/ambrose_er.html
- **toPath** = ecological-reserves/ambrose-lake

molund (changes requested):

molund (approved):

foley-lynn:
Natalia - can you add some comments to the ticket about what you've implemented, and also add some comments on this PR about how to test it?

I also think this would be a good time to update `main` on your local, and then run `rebase main` on this branch and force push. My understanding is that this PR relies on https://github.com/bcgov/bcparks.ca/pull/469, but this branch is behind that one, so I get a build error when I check it out locally.

Let Mike or I know if you need a hand with that.

ayumi-oxd (commented):
I tested `yarn build` with `DEV` env and got a build error below.

```
"gatsby-node.js" threw an error while running the
createPages lifecycle:

Invalid url.

  323 |   })
  324 |   parks.map(park => {
> 325 |     const oldUrl = parseUrl(park.oldUrl);
      |                    ^
  326 |     if(oldUrl.pathname !== `/${park.slug}/`) {
  327 |       return actions.createRedirect({
  328 |         fromPath: oldUrl.pathname,
```

It could be my environment but you might have this error when you merge it. Could you take a look at it? Thanks.

ayumi-oxd:
I think the build was not successful on my local because DEV, TEST, and PROD don't have a `parse-url` yet. Another build with my local Strapi worked so a new build on Github should be OK when you merge this PR.

ayumi-oxd (approved):

PR #476, opened 2022-12-16
NOBUG: Trying to fix GraphQL build error using createTypes definition

molund:
### Jira Ticket:
None

### Jira Ticket URL:
None

### Description:
Trying to fix GraphQL build error using createTypes definition (based on https://github.com/strapi/gatsby-source-strapi/issues/94)


foley-lynn (approved):

molund:
Closing because the solution doesn't fully work.  However, I do have a better understanding now of how to add a field without breaking graphQL.  You just need to add the new field to the typeDefs in gatsby-node.js

PR #485, opened 2022-12-20
CM-163: Add conditions for visible toggles

ayumi-oxd:
### Jira Ticket:
CM-163

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/CM-163

### Description:
- Added a fallback in case that `park.type` is null
- Added a visible condition for the Overview, the Operation, the Safety Info, and the Learn About section

### DIFF
--- a/src/staging/src/templates/park.js
+++ b/src/staging/src/templates/park.js
@@ -55,6 +55,7 @@ export default function ParkTemplate({ data }) {
   const apiBaseUrl = data.site.siteMetadata.apiURL
 
   const park = data.strapiProtectedArea
+  const parkType = park.type ?? "park"
   const photos = [...data.featuredPhotos.nodes, ...data.regularPhotos.nodes]
   const operations = park.parkOperation || {}
 
@@ -147,9 +148,9 @@ export default function ParkTemplate({ data }) {
   const menuItems = [
     {
       sectionIndex: 0,
-      display: capitalizeFirstLetter(`${park.type} overview`),
+      display: capitalizeFirstLetter(`${parkType} overview`),
       link: "#park-overview-container",
-      visible: true,
+      visible: park.description,
     },
     {
       sectionIndex: 1,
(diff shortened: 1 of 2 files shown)


foley-lynn (approved):

src/staging/src/templates/park.js
```diff
@@ -147,9 +148,9 @@ export default function ParkTemplate({ data }) {
   const menuItems = [
     {
       sectionIndex: 0,
-      display: capitalizeFirstLetter(`${park.type} overview`),
+      display: capitalizeFirstLetter(`${parkType} overview`),
       link: "#park-overview-container",
-      visible: true,
+      visible: park.description,
```
molund, line 153: There might be cases where an editor leaves behind a non-zero-length string or an empty `<p></p>` or `<br>` in the RTE, and it would break this condition check.  
You could create a function that uses a regex to remove tags and then trims the string.

molund:
```
// function to check if a string contains anything besides html tags and whitespace characters
const isNotWhiteSpace = (str) => str.replace(/(<([^>]+)>)|^\s+|\s+$|\s+/g, "").length > 0;
```

`visible: isNotWhiteSpace(park.description)`

ayumi-oxd: Thanks! I just copied yours as is 😂

molund: Sorry there was a bug in that version where it was choking on null.  Try the new one I posted.  Note that it's inverted now so you have to check for `!`
`!isNullOrWhiteSpace(park.description)`
`!isNullOrWhiteSpace(park.parkOperation)`
`!isNullOrWhiteSpace(park.safetyInfo)`
`!isNullOrWhiteSpace(park.specialNotes)`

ayumi-oxd: Thanks! I also applied this function to other `richtext` values, such as `park.parkContact`, `park.natureAndCulture`, and `park.reconciliationNotes`

molund (approved):

PR #488, opened 2022-12-21
CM-187: Update 404 page

ayumi-oxd:
### Jira Ticket:
CM-187

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/CM-187

### Description:
- Removed "Learn more" links
- Updated screenshots to reflect the current site
- Changed screenshots size from 1500px to 750px
- Added header and menu
- Added error message
- Updated links
- Code refactoring


src/staging/src/styles/404.scss
```diff
@@ -1,17 +1,20 @@
 @import "./global.scss";
 
-$header-height-404: 100px;
-$footer-height-404: 413px;
+$header-height-404: 220px;
+$footer-height-404: 390px;
```
foley-lynn, line 4: Ugh, I don't like this, but we didn't build it. I'd rather use flexbox properly than have to do this silly calculation. But this is ok for now.

src/staging/src/styles/404.scss
```diff
@@ -1,17 +1,20 @@
...
   color:white;
   margin-bottom: 0px;
   h1 {
     font-weight: 100;
   }
 
+  a {
+    color: white;
+  }
+
```
foley-lynn, line 17: I'd rather you use `class="text-white"` on the element in this case.

src/staging/src/pages/404.js
```diff
@@ -1,46 +1,42 @@
...
+              </p>
             </div>
             <div className="mt-2 d-flex justify-content-center">
               <div className="row">
                 <div className="col-lg-4 col-md-12 mt-2">
                   <ThumbLink imageLink="404/home-thumb.jpg" title="Homepage" navLink="/" />
                 </div>
                 <div className="col-lg-4 col-md-12 mt-2">
-                  <ThumbLink imageLink="404/park-thumb.jpg" title="Park Search" navLink="/find-a-park" />
+                  <ThumbLink imageLink="404/park-thumb.jpg" title="Find a park" navLink="/find-a-park" />
```
foley-lynn, line 36: The Thumblink component is unnecessarily fussy. If you're in the mood, I'd rather you just take it out completely since this is the only place we use it, and it's more complex than we need.

If that looks like it might cause issues, it's ok to leave it.

src/staging/src/pages/404.js
```diff
@@ -1,46 +1,42 @@
 import * as React from "react"
-import { Link } from "gatsby"
 
-import BetaBanner from "../components/betaBanner"
+import Header from "../components/header"
```
foley-lynn, line 3: On my local I'm getting a blank blue banner where the nav usually is, and on mobile, the hamburger menu is there, and opens a blank panel.

src/staging/src/pages/404.js
```diff
@@ -1,46 +1,42 @@
...
                 <div className="col-lg-4 col-md-12 mt-2">
                   <ThumbLink imageLink="404/home-thumb.jpg" title="Homepage" navLink="/" />
                 </div>
                 <div className="col-lg-4 col-md-12 mt-2">
-                  <ThumbLink imageLink="404/park-thumb.jpg" title="Park Search" navLink="/find-a-park" />
+                  <ThumbLink imageLink="404/park-thumb.jpg" title="Find a park" navLink="/find-a-park" />
                 </div>
                 <div className="col-lg-4 col-md-12 mt-2">
-                  <ThumbLink imageLink="404/reserve-thumb.jpg" title="Reservations" navLink="/reserve" />
+                  <ThumbLink imageLink="404/reserve-thumb.jpg" title="Reservations" navLink="/reservations" />
```
foley-lynn, line 39: This is optional, but these images are twice the size they need to be.

foley-lynn (changes requested):

src/staging/src/styles/404.scss
ayumi-oxd: Yes, this calculation looks hacky and it's not a sustainable way. I deleted it because it will unlikely be less than 100vh. Minimum contents (header, content, and footer) are more than 100vh on most of the devices.

src/staging/src/pages/404.js
ayumi-oxd: It might be useful to keep the `Thumblink` component since we can pass three different props. We could map the image element but assigning each link and title to each image element requires us to make some functions. So I'd prefer to keep it for the code readability. However, it's true that the `Thumblink` component is kinda fussy so I deleted unnecessary code.

src/staging/src/pages/404.js
ayumi-oxd: Good catch! I totally missed including the menu data...Thank you 🙏

src/staging/src/pages/404.js
foley-lynn: Makes sense, thanks!

foley-lynn (approved):

molund (approved):

PR #489, opened 2022-12-21
BRS-974: switch oidc to use loginproxy instead for bc parks admin portal

marklise:
### Jira Ticket:
BRS-974

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/BRS-974

### Description:
This switches the default local/dev setup to use the dev instance for the new KC Realm.

NB: There are a few ENV vars that will need to be updated after this is merged, prior to deployment into dev/running locally: `STRAPI_SSO_AUTH_URL` and `STRAPI_SSO_REALM`.  For environments in openshift, the auth url: `REACT_APP_KEYCLOAK_AUTH_URL` and `REACT_APP_KEYCLOAK_REALM` need changing from dev.oidc => dev.loginproxy (and similarly for test/prod). The realm in every environment is simply `bcparks-service-transformation`.


molund:
There are 2 additional settings in `/src/admin/.env.example` that also need to be updated

marklise:
> There are 2 additional settings in `/src/admin/.env.example` that also need to be updated. But they are currently incorrect, so we can update them later. Devs working locally on this project normally use the dev keycloak instance, not a local docker instance as indicated by the example file.

I've pushed an update so the next dev doesn't have to worry about this. :)

molund:
Thanks Mark.  We'll merge this first thing in January.  We won't want to risk breaking anything right before the holidays.

molund (approved):

ayumi-oxd (approved):

PR #492, opened 2022-12-22
CM-178: addressed seo issues with page and park templates 

foley-lynn:
### Jira Ticket:
CM-178

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/CM-178

### Description:
This is an attempt to fix some of the more egregious SEO errors on the page templates. Mainly has to do with meta tags.


src/staging/src/components/park/advisory.js
```diff
@@ -56,6 +56,7 @@ export default function Advisory({ advisories }) {
             src={alertIcon}
             aria-label="park access status"
             className="park-overview-icon"
+            alt="icon"
```
molund, line 59: Decorative images usually don’t need alt text.
I believe `alt=""` is actually better for screen readers.

src/staging/src/templates/staticLanding1.js
```diff
@@ -70,9 +70,9 @@ const LandingPage = ({ pageContext }) => {
   return (
     <>
       <Seo
-        title={meta.metaTitle}
-        description={meta.description}
-        keywords={meta.metaKeywords}
+        title={meta?.metaTitle || pageContext.page.Title}
```
molund, line 73: I think that should be page?.Title based on nearby examples

src/staging/src/components/park/advisory.js
foley-lynn: Thanks for catching this. This is a MUI component, and if you put `alt=""` in it strips it to `alt`, which is why I went with the above. I later discovered that from a screen-reader and SEO POV `alt=""` and `alt` are equivalent. So I'll switch all of those.

src/staging/src/templates/staticLanding1.js
foley-lynn: Thanks. page?.title wasn't working so I dug this up. I missed the capital 'T'.

molund (approved):

ayumi-oxd (commented):
I have some q's 🙂
- Do we need to have `seo` component for `site` page as the same as `park` page?
- Do we need to pull `seo` component out from the content list at `page` collection? To make it consistent with `protectedArea` and `parkSubPage` collections.

Other than that it looks good to me 👏
- I was able to see `seo` component on `protectedArea`
- I was able to build this PR with DEV, TEST, and PROD Strapi environment on my local

src/staging/gatsby-node.js
```diff
@@ -227,6 +228,7 @@ exports.createSchemaCustomization = ({ actions }) => {
     metaKeywords: String
     metaTitle: String
     metaDescription: String
+    canonical: String
```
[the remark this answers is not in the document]
foley-lynn: Good catch, I was actually implementing in incorrectly. this has been removed.

src/staging/src/templates/parkSubPage.js
```diff
@@ -80,7 +80,7 @@ export default function ParkSubPage({ data }) {
     <>
       <Seo
         title={seo?.metaTitle ?? park.protectedAreaName + ": " + page.title}
-        description={seo?.description}
+        description={seo?.metaDescription}
         keywords={seo?.metaKeywords}
       />
```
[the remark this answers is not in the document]
foley-lynn: No, I was implementing the canonical links incorrectly. With the plugin, they're on all pages automatically.

foley-lynn:
> * Do we need to have `seo` component for `site` page as the same as `park` page?

Yes, I've now added that.

> * Do we need to pull `seo` component out from the content list at `page` collection? To make it consistent with `protectedArea` and `parkSubPage` collections.

Probably, but it's messy. The whole way the pages are set up might just be better to not touch.

src/cms/api/protected-area/models/protected-area.settings.json
```diff
@@ -181,6 +181,11 @@
     "parkSubPages": {
       "via": "protectedArea",
       "collection": "park-sub-page"
+    },
+    "seo": {
+      "type": "component",
+      "repeatable": false,
+      "component": "parks.seo"
     }
```
ayumi-oxd, line 189: I think you'd need to add the same setting to `site.setting.json` to use `seo` component in `site.js`

src/staging/gatsby-node.js
```diff
@@ -104,6 +104,7 @@ exports.createSchemaCustomization = ({ actions }) => {
     parkFacilities: [StrapiParkFacilities]
     parkOperation: StrapiParkOperation
     parkOperationSubAreas: [StrapiParkOperationSubAreas]
+    seo: StrapiSeoComponent
   }
```
ayumi-oxd, line 108: I think you'd need to define type for `seo` at `site` collection. Adding `seo: StrapiSeoComponent` to `type StrapiSites implements Node` should work.

ayumi-oxd (commented):
Added a few comments, but other than that it looks great 👏

src/cms/api/protected-area/models/protected-area.settings.json
foley-lynn: I'm using the component in the front-end, but not using any of the fields in site pages. For now it's just using the park and site name for the title, and truncating the description for the meta description.

We should integrated the SEO component in eventually, but not for MVP.

src/staging/gatsby-node.js
foley-lynn: As above, for sites we're just using the component to populate the title and the description based on existing fields. So I think it's ok to not register it here. I might be wrong though, so if you know better, let me know.

ayumi-oxd (approved):
Sorry I didn't notice `site.js` doesn't use data from `Seo`. Works great as is 👍

PR #499, opened 2022-12-28
CM-224: Resolve console errors on Park pages

nataliasabadysh:
### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/CM-224

### Description:
1) useEffect should have a correct dependency 
2) we don't need to use Promise.all for one API request, for all side effects useEffect will take care of it, even if we need to chain them


foley-lynn:
First error is cleared, still getting the second one:
`Warning: Failed prop type: Invalid prop `parkName` of type `object` supplied to `ParkHeader`, expected `string`.`

ayumi-oxd (approved):
LGTM 👏

foley-lynn (approved):
Looks great.

PR #505, opened 2023-01-04
CM-279: Change api endpoint for counter

ayumi-oxd:
### Jira Ticket:
CM-279

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/CM-279

### Description:
- There are 455 `PublicAdvisories` on [TEST](https://test-cms.bcparks.ca/admin/plugins/content-manager/collectionType/application::public-advisory.public-advisory?page=1&pageSize=10&_sort=title:ASC)
- [Previous endpoint](https://test-cms.bcparks.ca/public-advisories/count?protectedAreas.published_at_null=false&protectedAreas.isDisplayed=true) `/public-advisories/count?protectedAreas.published_at_null=false&protectedAreas.isDisplayed=true` has 550 `PublicAdvisories`
- Changed endpoint to `/public-advisories/count` to get proper data
- ^ [New endpoint](https://test-cms.bcparks.ca/public-advisories/count?) has 455 `PublicAdvisories`
#### Issues:
- Somehow filtered endpoint count doesn't work properly 🥲
- However, filtered endpoint and additional query works (e.g. [Example with search query](https://test-cms.bcparks.ca/public-advisories/count?_sort=advisoryDate:DESC&protectedAreas.published_at_null=false&protectedAreas.isDisplayed=true&_q=Sproat) has 2 `PublicAdvisories`)

### DIFF
--- a/src/staging/src/pages/active-advisories.js
+++ b/src/staging/src/pages/active-advisories.js
@@ -161,10 +161,10 @@ const PublicActiveAdvisoriesPage = ({ data }) => {
 
     // exclude unpublished parks
     let q =
-      "/public-advisories/count?protectedAreas.published_at_null=false&protectedAreas.isDisplayed=true"
+      "/public-advisories/count"
 
     if (advisoryType !== "all") {
-      q += `&eventType.eventType_contains=${advisoryType}`
+      q += `?protectedAreas.published_at_null=false&protectedAreas.isDisplayed=true&eventType.eventType_contains=${advisoryType}`
     }
 
     const newApiCountCall = apiUrl + q
@@ -250,6 +250,10 @@ const PublicActiveAdvisoriesPage = ({ data }) => {
 
             // Get count
             let apiCount = apiUrl + "/public-advisories/count" + q
+            if (q === "?_sort=advisoryDate:DESC&protectedAreas.published_at_null=false&protectedAreas.isDisplayed=true") {
+             apiCount = apiUrl + "/public-advisories/count"
+            }
+            
             axios
               .get(apiCount)
               .then(function (data) {
--- a/src/staging/src/pages/active-advisories.js
+++ b/src/staging/src/pages/active-advisories.js
@@ -160,11 +160,15 @@ const PublicActiveAdvisoriesPage = ({ data }) => {
     // unfiltered count for the header
 
     // exclude unpublished parks
+    // this filter has been removed as a temporary workaround for a Strapi bug.
+    // see https://github.com/bcgov/bcparks.ca/pull/505/files#r1067160153
+
+    // let q = "/public-advisories/count?protectedAreas.published_at_null=false&protectedAreas.isDisplayed=true"
     let q =
       "/public-advisories/count"
 
     if (advisoryType !== "all") {
-      q += `?protectedAreas.published_at_null=false&protectedAreas.isDisplayed=true&eventType.eventType_contains=${advisoryType}`
+      q += `?&eventType.eventType_contains=${advisoryType}`
     }
 
     const newApiCountCall = apiUrl + q


src/staging/src/pages/active-advisories.js
```diff
@@ -161,10 +161,10 @@ const PublicActiveAdvisoriesPage = ({ data }) => {
 
     // exclude unpublished parks
     let q =
-      "/public-advisories/count?protectedAreas.published_at_null=false&protectedAreas.isDisplayed=true"
+      "/public-advisories/count"
 
     if (advisoryType !== "all") {
-      q += `&eventType.eventType_contains=${advisoryType}`
+      q += `?protectedAreas.published_at_null=false&protectedAreas.isDisplayed=true&eventType.eventType_contains=${advisoryType}`
     }
```
molund, line 168: Is this logic correct? It seems like you've removed the filter for the "all" advisory type, so it will include unpublished advisories and advisories from undisplayed protected areas?

ayumi-oxd: I'm not sure why, but without filtering has the correct number of counts
- Without filtering, `https://test-cms.bcparks.ca/public-advisories/count?` has 455 which is correct as you can see [here](https://test-cms.bcparks.ca/admin/plugins/content-manager/collectionType/application::public-advisory.public-advisory?page=1&pageSize=10&_sort=title:ASC)
- With filtering, `https://test-cms.bcparks.ca/public-advisories/count?protectedAreas.published_at_null=false&protectedAreas.isDisplayed=true` has 550 which is incorrect
- With filtering, and additional filter or query, `https://test-cms.bcparks.ca/public-advisories/count?protectedAreas.published_at_null=false&protectedAreas.isDisplayed=true&eventType.eventType_contains=anchorage` has 3 which is correct as you can see [here](https://test-cms.bcparks.ca/admin/plugins/content-manager/collectionType/application::public-advisory.public-advisory?page=1&pageSize=10&_sort=eventType.eventType:ASC&_where[0][eventType.eventType]=Anchorage)

...but I couldn't find why the counter with filtering doesn't work properly 🥲 Please let me know if you find anything.

molund: The root cause is this Strapi bug...
https://github.com/strapi/strapi/issues/7631
Right now, we don't have any protectedAreas with isDisplayed = false (or any unpublished protected areas), so the code is working with no filter.  But if we had public advisories for these protectedAreas, then it would return the wrong count. 

I think we have 2 options:
1. Remove the deep filtering (as you have done) and use the unfiltered count.
2. Add a custom count endpoint to the protected-areas controller that uses a knex subquery to get the correct count.

Option 1 is only an issue if we have unpublished protectedAreas that also have publicAdvisories.

Option 2 fixes the issue for Strapi 3, but then we are adding a custom endpoint that probably won't be needed for Strapi 4.

src/staging/src/pages/active-advisories.js
```diff
@@ -161,10 +161,10 @@ const PublicActiveAdvisoriesPage = ({ data }) => {
 
     // exclude unpublished parks
     let q =
-      "/public-advisories/count?protectedAreas.published_at_null=false&protectedAreas.isDisplayed=true"
+      "/public-advisories/count"
 
     if (advisoryType !== "all") {
-      q += `&eventType.eventType_contains=${advisoryType}`
+      q += `?protectedAreas.published_at_null=false&protectedAreas.isDisplayed=true&eventType.eventType_contains=${advisoryType}`
```
molund, line 167: I think this line should be reverted based on our new understanding of the underlying issue.  Then I'll approve it with option 1 from my comment below.

src/staging/src/pages/active-advisories.js
```diff
@@ -161,10 +161,10 @@ const PublicActiveAdvisoriesPage = ({ data }) => {
 
     // exclude unpublished parks
     let q =
-      "/public-advisories/count?protectedAreas.published_at_null=false&protectedAreas.isDisplayed=true"
```
molund, line 164: instead of removing this line, can you comment it out and also add a code comment with a link to my comment from below...
https://github.com/bcgov/bcparks.ca/pull/505/files#r1067160153

molund (approved):

foley-lynn (approved):

PR #511, opened 2023-01-06
NOBUG: Add validation and error message for slugs

ayumi-oxd:
### Jira Ticket:
Relates to CM-314 / CM-318 

### Jira Ticket URL:
Relates to https://bcparksdigital.atlassian.net/browse/CM-314 / https://bcparksdigital.atlassian.net/browse/CM-318

### Description:
*This PR should be merged after MVP release*
- Added a slug validation `^[a-z0-9\-\/]+(?:-[a-z0-9\-\/]+)*$|^$` to allow editors to enter lowercase letters, numbers, hyphens, or slashes, but no spaces, based on [this doc](https://apps.nrs.gov.bc.ca/int/confluence/display/BCPRS/URLs%2C+menus%2C+and+links)
- Added this validation to `Pages`, `ParkSubPages`, `ProtectedAreas`, and `Sites`
- Added a custom error message to let editors know

Screen Shot 2023-01-05 at 5 56 42 PM (2)


src/cms/api/park-sub-page/models/park-sub-page.js
```diff
@@ -5,4 +5,20 @@
  * to customize this model
  */
 
-module.exports = {};
+const regex = new RegExp("^[a-z0-9/]+(?:-[a-z0-9/]+)*$")
```
molund, line 8: Is this one intentionally different from the other?

ayumi-oxd: Yes, it's because the slug field in `parkSubPage` is required. So it shouldn't allow an empty string.

src/cms/api/site/models/site.js
```diff
@@ -5,4 +5,20 @@
  * to customize this model
  */
 
-module.exports = {};
+const regex = new RegExp("^[a-z0-9/]+(?:-[a-z0-9/]+)*$|^$");
+const slugValidator = (data) => {
```
molund, line 9: Can we create a shared file instead of having the same code in 4 places?

ayumi-oxd: I thought the same thing but faced [this problem](https://forum.strapi.io/t/how-to-structure-reusable-code-to-import-in-various-strapi-files/2435/3).

- Created `utils/helpers.js` under `cms`
- In `helpers.js`
```
const regex = new RegExp("^[a-z0-9/]+(?:-[a-z0-9/]+)*$|^$");
export const slugValidator = (slug) => {
    if (!regex.test(slug)) {
      throw strapi.errors.badRequest('Please enter letters, numbers, hyphens, or slashes for slug. No spaces.');
    }
};
```
- In `page.js`
```
import { slugValidator } from "../../../utils/helpers.js";
```
- Got an error: `SyntaxError: Cannot use import statement outside a module`

Seems like I could use [this configuration](https://docs.strapi.io/developer-docs/latest/setup-deployment-guides/configurations.html#functions) but not sure what to return.

Do you have any ideas?

molund: It's actually server-side node code not client-side javascript code.  So I believe you need to use `require` instead of `import`

`const slugValidator = require("../../../utils/helpers.js").slugValidator();`

(I think it's something like that anyway)

src/cms/config/functions/helpers.js
```diff
@@ -0,0 +1,6 @@
+module.exports = function (slug) {
+  const regex = new RegExp("^[a-z0-9\-\/]+(?:-[a-z0-9\-\/]+)*$|^$");
+    if (!regex.test(slug)) {
+      throw strapi.errors.badRequest('Please enter lower case letters, numbers, hyphens, or slashes for slug. No spaces.');
+    }
+}
```
molund, line 6: We should call this something more descriptive, since it only contains one function.

e.g. slugValidator.js

molund (changes requested):

src/cms/api/page/models/page.js
```diff
@@ -5,4 +5,15 @@
  * to customize this model
  */
 
-module.exports = {};
+const slugValidator = require("../../../config/functions/slugValidator.js");
```
molund, line 8: Slugs for pages should always begin with a forward slash.  The regex allows me to enter a slug without the forward slash.

Is there something we can do to make it less brittle?  for example, chaining adding a couple of replace() statements to gatsby-node.js will ensure there is always a beginning and ending slash even if the user didn't enter one....
```
      actions.createPage({
        path: page.Slug.replace(/^\/|^/, `/`).replace(/\/$|$/, `/`),
        component: require.resolve(`./src/templates/${page.Template}.js`),
        context: { page },
      })
```

src/cms/config/functions/slugValidator.js
```diff
@@ -0,0 +1,6 @@
+module.exports = function (slug) {
+  const regex = new RegExp("^[a-z0-9\-\/]+(?:-[a-z0-9\-\/]+)*$|^$");
+    if (!regex.test(slug)) {
+      throw strapi.errors.badRequest('Please enter lower case letters, numbers, hyphens, or slashes for slug. No spaces.');
+    }
```
molund, line 5: can we have a second test to prevent consecutive forward slashes

e.g. 
```
  const regex2 = new RegExp(/^(?![\w/-]*\/\/)[\w/-]+$/gm);
  if (!regex2.test(slug)) {
    throw strapi.errors.badRequest('Consecutive forward slashes are not allowed in slugs.');
  }
```

molund:
maybe we could a bunch of different slug validators with different regex expressions and stack them for the particular field based on it's unuque circumstances?

1. slugCharacterValidator - checks for valid characters and consecutive forward slashes (the current validator plus my comment above)
2. slugLeadingSlashValidator - a leading slash is required
3. slugNoLeadingSlashValidator - a leading slash is not allowed
4. slugTrailingSlashValidator - a trailing slash is required
5. slugNoTrailSlashValidator - a trailing slash is not allowed 
6. slugNoSlashValidator - foward slashes are not allowed in the slug

you would stack them like this.....

```
    beforeUpdate: async (params, data) => {
      slugCharacterValidator(data.Slug)
      slugNoSlashValidator(data.Slug)
    },
```

molund:
This is looking good, but I noticed a couple of small bugs while testing.
1. If I leave the slug blank on a Site, I get the error about consecutive forward slashes.  I believe sites should still allow blank slugs?

2.  We probably need to add better handling for nulls.  I was getting this error in my console 
**[2023-01-18T17:58:58.111Z] error TypeError: Cannot read property 'slice' of undefined**

ayumi-oxd:
- Changed `regex2` to `^(?!.*\/\/)[a-z0-9\-\/]+(?:-[a-z0-9\-\/]+)*$|^$`, which allows editors to keep slug empty on `Sites` page

molund:
Can you make one more small change please, and set the slug to be required for pages and protected-areas?  
image

molund (approved):

foley-lynn:
My only note is that the error message disappears quickly. I'd triple the time before it goes away, or just leave it if it has a working 'x' and it's not covering other things.

Let's be careful about re-using this, as it's pretty strict. Sometimes we'll need '.' or '#' in the slug.

In this case, it looks good to me.

foley-lynn (approved):
Approved with one comment.

ayumi-oxd:
@foley-lynn Yes, that's a shame that the error message fades quickly. Unfortunately, I couldn't find an easy solution for that since it's a part of the Strapi lifecycle hooks. This validation will need to be changed after updating v4 anyway, so I think it's OK not to take much time for this at this moment.

FYI 🙂
- [Strapi notification](https://github.com/strapi/strapi/blob/releases/3.6.11/packages/strapi-admin/admin/src/components/Notifications/NotificationsContainer/index.js) - It fades after 3 sec 😅
- [Throw error message in lifecycle](https://forum.strapi.io/t/throw-error-message-in-lifecycle-hook-doesnt-change-the-error-message-display-in-the-ui/1515) - There is a difference between v3 and v4 lifecycle hooks, so we need to update this again...

foley-lynn:
Sounds good, thanks for looking.

PR #524, opened 2023-01-12
CM-244: Format the park operations section

molund:
### Jira Ticket:
CM-244

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/CM-244

### Description:
Format the "Dates of operation" section on Parks pages according to wireframes in the ticket

### DIFF
--- a/src/staging/src/components/park/parkDates.js
+++ b/src/staging/src/components/park/parkDates.js
@@ -1,5 +1,6 @@
 import React, { useState } from "react"
 import moment from "moment"
+import _ from "lodash"
 
 import HTMLArea from "../HTMLArea"
 import Row from "react-bootstrap/Row"
@@ -34,11 +35,9 @@ export default function ParkDates({ data }) {
   // If no operations record, show "not available" message
   const hasOperations = parkOperation.isActive // either false, or whole record missing
 
-  const fmt = "MMM D, yyyy" // date format for display
-
   // -------- Operating Dates --------
 
-  const datePhrase = (openDate, closeDate) => {
+  const datePhrase = (openDate, closeDate, fmt, yr) => {
     if (openDate && closeDate) {
       try {
         const open = moment(openDate).format(fmt)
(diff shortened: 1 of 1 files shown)


foley-lynn (approved):

foley-lynn:
This looks really good Mike. Nice work.

ayumi-oxd:
One question - I was at `/adams-lake-bush-creek-site-park/` with PROD data. I see "Open to public access May 13, 2022 to September 15, 2022" under the Dates of operation, but it's supposed to be not displayed based on Jane's notes in the wireframe. Sorry if I missed something 😅

Other than that it looks great 👏

molund:
@ayumi-oxd 
I'm was expecting that once I refresh the park operation data with 2023 information, the dates for Adams Lake Bush Creek Site would be 2023 dates.  However, I checked prod and I noticed that this park currently has no operating dates section at all.....
https://bcparks.ca/explore/parkpgs/adams_lk_bush_crk/
I wonder if maybe the preferred fallback behaviour in January is to hide the entire "Dates of Operation" section instead of showing last year's dates?

ayumi-oxd:
@molund That fallback sounds right to me 🙂 But also I'm concerned we wouldn't have the Dates of operation on any park pages until `ParkOperations` data is updated from 2022 to 2023. At this moment, I don't see [any data that starts or ends in 2023](https://cms.bcparks.ca/admin/plugins/content-manager/collectionType/application::park-operation.park-operation?page=1&pageSize=10&_sort=openDate:DESC). Talking to Jane about this might be a good idea.

molund:
@ayumi-oxd I asked Leah, and she wanted to change the fallback.  Now if the dates are missing we will still show the section, but instead of dates it will say "Operating dates are not yet available".  In addition, the accordion sections will now appear when there are no dates as well.  But under "Main season camping dates" it will say "2023: Dates are not yet available".
image (3)

ayumi-oxd (approved):
I was able to check the new fallback on my local, and it looks awesome 👏

ayumi-oxd:
@molund Sorry one nitpicking 😅 If you could change `class` to `className`, it would be more awesome!

molund:
Done.  Good catch!  Can you tell they never let me do front-end work in React? 😄

PR #530, opened 2023-01-14
CM-90: Update side nav behaviour

ayumi-oxd:
### Jira Ticket:
CM-90

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/CM-90

### Description:
- Added height to the last section of the park page/site page to apply the active link style to the last side nav link
- Changed `offsetPx` to apply the active link style to the section that has a minimum height
- Deleted unused refs

### DIFF
--- a/src/staging/src/styles/parks.scss
+++ b/src/staging/src/styles/parks.scss
@@ -2,13 +2,17 @@
 @import url("https://js.arcgis.com/4.10/esri/css/main.css");
 // @import url("https://governmentofbc.maps.arcgis.com/apps/webappviewer/jimu.js/css/jimu-theme.css");
 
-$mediumBreakpoint: 991.98px;
-
 .parks-container {
   //Added 50px extra to compensate padding
   max-width: 1370px !important;
   font-size: 15px;
   padding: 0px !important;
+  &.main-container {
+    padding-bottom: 0px !important;
+    @media (min-width: $lgBreakpoint) {
+      padding-bottom: 250px !important;
+    }
+  }
 }
 
 .anchor-linkx {
--- a/src/staging/src/styles/parks.scss
+++ b/src/staging/src/styles/parks.scss
@@ -7,11 +7,8 @@
   max-width: 1370px !important;
   font-size: 15px;
   padding: 0px !important;
-  &.main-container {
-    padding-bottom: 0px !important;
-    @media (min-width: $lgBreakpoint) {
-      padding-bottom: 250px !important;
-    }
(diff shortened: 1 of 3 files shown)


src/staging/src/styles/parks.scss
```diff
@@ -2,13 +2,17 @@
...
   //Added 50px extra to compensate padding
   max-width: 1370px !important;
   font-size: 15px;
   padding: 0px !important;
+  &.main-container {
+    padding-bottom: 0px !important;
+    @media (min-width: $lgBreakpoint) {
+      padding-bottom: 250px !important;
+    }
+  }
```
foley-lynn, line 15: The real issue is that the last section needs to be a certain height to trigger the scroll spy.

So I'd do something like: 
`.makeStyles-parkContent-2 > div:last-of-type {
    min-height: 600px;
}`

Though you may want to put a unique class on to that container as I don't know if `.makeStyles-parkContent-2` is unique.

But I'll also note that this issue exists on content pages as well, and I'll note that there seems to be an empty div at the end of some (all?) of the `.page-content` parent divs.

foley-lynn (changes requested):

src/staging/src/styles/parks.scss
foley-lynn: FYI - I chose 600px as that'll trigger the final title if the display is 1080px high. That'll cover 97%+ of the cases. More looks goofy.

ayumi-oxd: Thanks! `.makeStyles-parkContent-2 > div:last-of-type { min-height: 600px; }` works perfectly 👏
I checked `.makeStyles-parkContent-2` is used in `park.js` and `site.js`, and it's okay since both templates are the same layout.

For the content page, let me work on the same issue in [this ticket]( https://bcparksdigital.atlassian.net/browse/CM-91) 🙂

src/staging/src/templates/park.js
```diff
@@ -341,7 +342,7 @@ export default function ParkTemplate({ data }) {
               sm={12}
               md={9}
               lg={9}
-              className={classes.parkContent}
+              className={`${classes.parkContent} test`}
```
foley-lynn, line 345: Committed comment. Also the console.log above.

foley-lynn (changes requested):
Looks good, just scrub those two changes from park.js

ayumi-oxd:
Totally forgot about deleting them 😅 Thanks!

foley-lynn (approved):

molund (approved):

PR #542, opened 2023-01-19
Cm 289 : Add ID attribute to iframe code in source view

nataliasabadysh:
### Jira Ticket URL:
https://bcparksdigital.atlassian.net/jira/software/projects/CM/boards/11?assignee=633db6bb61dbef2805c1f25c&selectedIssue=CM-289

### DIFF
--- a/src/staging/src/hooks/useContent.js
+++ b/src/staging/src/hooks/useContent.js
@@ -1,0 +1,46 @@
+import { useState, useEffect } from "react";
+import * as cheerio from "cheerio";
+
+export const useContent = (contentHtml) => {
+  const { introHtml } = contentHtml;
+  const [title, setTitle] = useState("");
+
+  let $ = cheerio.load(introHtml);
+
+  let getIframe = $("iframe");
+  let media = $(".media");
+
+  const getURL = getIframe.attr("src");
+
+  const srcId = getURL?.split("/");
+  const id_end = srcId && srcId[srcId?.length - 1];
+
+  useEffect(() => {
+    if (id_end) {
+      const vidurl = `https://www.youtube.com/watch?v=${id_end}`;
+
+      fetch(`https://noembed.com/embed?dataType=json&url=${vidurl}`)
+        .then((response) => response.json())
+        .then((data) => {
+            setTitle(data.title);
+        })
+        .catch((error) => console.error(error));
+    }
+  }, [id_end]);
+
+  if (title) {
+    getIframe.attr("id", title);
+  }
+
+  const start = introHtml.indexOf("<figure");
+  const end = introHtml.indexOf("</figure>");
+
+  const videoMarkup = introHtml.slice(start, end);
+
+  const content =
+    start >= 0 &&
+    end >= 0 &&
+    introHtml.replace(videoMarkup, media.html());
+
+  return { introHtml : content };
+};
--- a/src/staging/package.json
+++ b/src/staging/package.json
@@ -16,6 +16,7 @@
     "@material-ui/styles": "^4.11.4",
     "axios": "^0.22.0",
     "bootstrap": "~4.3.1",
+    "cheerio": "^1.0.0-rc.12",
     "coffee-register": "~2.2.0",
     "date-fns": "^2.28.0",
     "gatsby": "^3.13.1",
--- a/src/staging/src/templates/staticContent1.js
+++ b/src/staging/src/templates/staticContent1.js
@@ -1,18 +1,20 @@
-import React, { useRef } from "react"
-import { graphql, useStaticQuery } from "gatsby"
-import { Breadcrumbs } from "@material-ui/core"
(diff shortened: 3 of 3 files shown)


molund:
Can you please add a description of the final solution to the PR and the ticket?

It looks like this was an issue with the Strapi RTE, but all the code changes are in Gatsby? Did you end up creating an ID to solve the analytics problem via Gatsby?

nataliasabadysh:
I left some comments on the ticket 
 
after trying CKEditor  I come up with this solution that worked for me 

from Strapi CKEditor I could not add this ID 
it is not allowed to async make requests to get a title for the video 

I created a hooks folder, where I moved all the logic from the component 
takes content and returns  it with video ID,  if that is needed 

` const { introHtml } = useContent(contentHtml)`


I added lib for the parsing HTML Content string 
https://cheerio.js.org/  - to add ID before it the render

in this lib, I could get specific html that I need by class, tag 

```
  let getIframe = $("iframe");
  let media = $(".media");
```
  
  the I add  ID=title ,  
  
`  getIframe.attr("id", title);`
   
   
  and then I replace the iframe in this Content string before returning it 
  
  if you have any comments please  let me know

foley-lynn:
@nataliasabadysh  - This fails for me when I try to build gatsby locally. It fails against local strapi and TEST strapi:

`  3 |
  4 | export const useContent = (contentHtml) => {
> 5 |   const { introHtml } = contentHtml;
    |           ^
  6 |   const [title, setTitle] = useState("");
  7 |
  8 |   let $ = cheerio.load(introHtml);


  WebpackError: TypeError: Cannot destructure property 'introHtml' of 'contentHtml' as it is undefined.`

molund:
```
Building static HTML failed for path "/training/invasive-species/"

See our docs page for more info on this error: https://gatsby.dev/debug-html


  3 |
  4 | export const useContent = (contentHtml) => {
> 5 |   const { introHtml={} } = contentHtml;
    |           ^
  6 |   const [title, setTitle] = useState("");
  7 |   let $ = cheerio.load(introHtml);
  8 |


  WebpackError: TypeError: Cannot read property 'introHtml' of undefined
```

I have my .env file set to use prod, and I am getting this error from `yarn build`

nataliasabadysh:
@molund sorry , I have that destructurization that should set default when there is no value,  
but I was not getting this error, I build it many times while working on it hmm  

I will update this in 2 min
just running locally

molund:
I confirmed that the build completed with no errors, and an ID matching the video title was added to 
http://localhost:9000/about/careers/park-ranger/

src/staging/src/hooks/useContent.js
```diff
@@ -0,0 +1,53 @@
...
+
+export const useContent = (contentHtml) => {
+  const [title, setTitle] = useState("");
+  let video_id;
+  let media; 
+  let getIframe
+
+  if(contentHtml?.introHtml){
+    let $ =  cheerio.load(contentHtml?.introHtml);
+    getIframe = $("iframe");
```
molund, line 12: I'm not sure if there are parks pages with multiple videos, but this selector could return more than 1 iframe.  
jQuery would return multiple matches if your selector had multiple matches.  It looks like Cheerio has a toArray() method that behaves the same as jQuery.  
https://www.tabnine.com/code/javascript/functions/cheerio/Cheerio/toArray

src/staging/src/hooks/useContent.js
```diff
@@ -0,0 +1,53 @@
...
+    media = $(".media");
```
molund, line 13: If this works like jQuery, then I think this selector could be more specific and use `figure.media` instead of `.media`

Also the media is a child/descendant of the iframe.  Is there a child selector you can use from the iframe, so you don't accidentally get a different `figure.media` from somewhere else on the page?

nataliasabadysh: @molund 
thank you for testing, helping to find and giving solutions! 
in any way better to add those cases, I will add  

1) Multiple options toAraay 
2) Specify elem by `figure.media`

ayumi-oxd (commented):
- Got a warning below after `yarn start` 
```
warn 
/Users/ayumi/Downloads/projects/bcparks/bcparks.ca/src/staging/src/hooks/useContent.js
13:11  warning  Assignments to the '$' variable from inside React Hook useEffect will be lost after each render. To preserve the value over time, store it in a useRef Hook and keep the mutable value in the '.current' property. Otherwise, you can move this variable directly inside useEffect
```

Other than that, looks good to me 🙂
- I was able to see the id in the `<figure>` after the build
- `yarn build` finished successfully

foley-lynn:
Looking good, but a few issues.

First off - the id is being added to `figure.media` but it should be added to the iFrame.

Secondly - the title should slugified and should have a max length. Let's say 80 characters.

molund:
Replaced by https://github.com/bcgov/bcparks.ca/pull/671

PR #545, opened 2023-01-19
CM-263: Exclude advisories for other sites at the same park

molund:
### Jira Ticket:
CM-263

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/CM-263

### Description:
On site pages, we only want to show advisories that are either 
a. At the park level or higher (they don't specify any specific sites)
b. That specify this specific site.
If an advisory specifies a specific site, but it is not the site for the current page, then it should be excluded.

### DIFF
--- a/src/staging/src/templates/site.js
+++ b/src/staging/src/templates/site.js
@@ -107,7 +107,17 @@ export default function ParkTemplate({ data }) {
     loadAdvisories(apiBaseUrl, park?.orcs)
       .then(response => {
         if (response.status === 200) {
-          setAdvisories([...response.data])
+          // for sites, we want to include all advisories at the park level  
+          // and advisories for this specific site, but we exclude advisories 
+          // for other sites at the same park
+          const advisories = [...response.data].filter(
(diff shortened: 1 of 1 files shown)


ayumi-oxd (approved):

ayumi-oxd:
Sorry, one nit 😅 You probably need to add `site.orcsSiteNumber` as a `useEffect` dependency. 

Warning:
```
warning  React Hook useEffect has a missing dependency:
'site.orcsSiteNumber'. Either include it or remove the dependency array
```
So adding a dependency at line 130:
```
  }, [apiBaseUrl, park?.orcs, site.orcsSiteNumber])
```
Hope it works 🙂

molund:
> ```
> warning  React Hook useEffect has a missing dependency:
> 'site.orcsSiteNumber'. Either include it or remove the dependency array
> ```

Good catch @ayumi-oxd .  I was using `yarn build` instead of `yarn start` and there was no warning

nataliasabadysh (approved):

PR #552, opened 2023-01-20
CM-387: Updates to token authentication

molund:
### Jira Ticket:
CM-387

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/CM-387

### Description:
This change is a prerequisite for CM-14 (loading park operating dates in Strapi)

Token authentication was broken and not working at all. 

I changed the implementation of API authentication using the "Tokens" collection type:
- It's no longer possible to pass a token via request.body or request.query, it must be passed as an authorization header (passing this as a querystring is a security risk because the token is very likely to end up in logs. And passing it as a request.body parameter seems like an unnecessary complication that would never be used anyway)
  - Note that passing tokens as via request.body or request.query was broken, so I am confident that nobody was currently using it.
- When passing a token via the authorization header it doesn't need to be a jwt token. It's just a plain string, but the authorization header value must start with "Bearer "

Note: when testing the staff portal with Keycloak auth, make sure your keycloak settings in both .env files are correct....

**admin**
```
REACT_APP_KEYCLOAK_AUTH_URL=https://dev.loginproxy.gov.bc.ca/auth
REACT_APP_KEYCLOAK_REALM=bcparks-service-transformation
REACT_APP_KEYCLOAK_CLIENT_ID=staff-portal
```

**cms**
```
STRAPI_SSO_AUTH_URL=https://dev.loginproxy.gov.bc.ca/auth
STRAPI_SSO_REALM=bcparks-service-transformation
```

### DIFF
--- a/src/cms/api/token/models/token.settings.json
+++ b/src/cms/api/token/models/token.settings.json
@@ -14,7 +14,10 @@
     "token": {
       "type": "string",
       "required": true,
-      "unique": true
+      "unique": true,
+      "minLength": 40,
+      "regex": "^[^\\s]+(\\s+[^\\s]+)*$",
+      "private": true
     },
     "user": {
       "unique": true,
--- a/src/cms/extensions/users-permissions/config/policies/permissions.js
+++ b/src/cms/extensions/users-permissions/config/policies/permissions.js
@@ -12,94 +12,68 @@ module.exports = async (ctx, next) => {
   }
 
   // add the detection of `token` query parameter
-  if (
-    (ctx.request && ctx.request.header && ctx.request.header.authorization) ||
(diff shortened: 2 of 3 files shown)


ayumi-oxd:
Looks good 🙂 I was able to test it using Postman and got data with the PUT method.

Screen Shot 2023-02-02 at 2 35 17 PM (2)

ayumi-oxd (approved):

molund:
Strapi 4 has API tokens built-in and it's better than the current token authentication mechanism in several ways.  I'm going to close this PR and we can use the Strapi 4 mechanism instead.  
- it uses real JWT tokens
- tokens are salted
- tokens can be viewed only once when they are generated

PR #573, opened 2023-02-06
CM-377: Sort activity names

ayumi-oxd:
### Jira Ticket:
CM-377

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/CM-377

### Description:
- Added `orderBy` to sort `activityName` by alphabet case insensitive order

### DIFF
--- a/src/staging/src/pages/find-a-park.js
+++ b/src/staging/src/pages/find-a-park.js
@@ -482,9 +482,6 @@ export default function FindAPark({ location, data }) {
 
  const getParkName = item => item.parkNameType === PARK_NAME_TYPE.Escaped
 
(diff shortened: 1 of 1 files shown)


molund:
The items are already being sorted on line 54.  
If the sort on line 54 can't be made case-insensitive then it should be deleted.

ayumi-oxd:
> The items are already being sorted on line 54. If the sort on line 54 can't be made case-insensitive then it should be deleted.

Good catch! Yes, GraphQL doesn't have a case-insensitive sort function 😔

molund (approved):
Works as expected

foley-lynn (approved):

PR #575, opened 2023-02-06
CM-97: Fix aria menu error

ayumi-oxd:
### Jira Ticket:
CM-97

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/CM-97

### Description:
- The error was because the element with `role="menu"` did not contain any child element with `role="menuitem"`
- Added `role="menuitem"` to the menu item elements
- Added `role="presentation"` to the  `<li>` elements 
- Ref: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role#example_1_navigation_menu


molund (approved):
Confirmed that this builds correctly and the mega-nav is still tab-able.  I'll leave it to QA to do extensive testing.

foley-lynn (approved):

PR #609, opened 2023-03-10
CM-529: Skip park overview by tabbing

ayumi-oxd:
### Jira Ticket:
CM-529

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/CM-529

### Description:
- Added `tabindex` to skip `<a>` element by tabbing when section is not expanded

### DIFF
--- a/src/staging/src/components/park/parkOverview.js
+++ b/src/staging/src/components/park/parkOverview.js
@@ -37,6 +37,7 @@ export default function ParkOverview({ data: parkOverview, type }) {
   }, [expanded])
 
   const isLong = height >= 260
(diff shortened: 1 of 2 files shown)


src/staging/src/components/park/parkOverview.js
```diff
@@ -37,6 +37,7 @@ export default function ParkOverview({ data: parkOverview, type }) {
   }, [expanded])
 
   const isLong = height >= 260
+  const collapsedParkOverview = parkOverview.split('<a').join('<a tabindex="-1"')
```
molund, line 40: This would also split on <abbr, <address, <article, <aside, <area.  Can you use something like **cheerio** to do this instead?

molund (approved):

PR #612, opened 2023-03-17
CM-544: slug validators should pass for blank/null/undefined slug

molund:
### Jira Ticket:
CM-544

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/CM-544

### Description:
When a brand-new page is published for the first time, the hooks are called, but the context is missing so the slug is treated as blank or undefined.  This PR updates the slug validators so they will no longer simulaneously act as required field validators.


ayumi-oxd (approved):

manuji (approved):
Looks good

PR #643, opened 2023-04-21
CM-596: Fix for advisory relations not saving

molund:
### Jira Ticket:
CM-596

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/CM-596

### Description:
Fixed an issue where relations weren't being saved to public advisories in the lifecycle hook on public advisory audits.


ayumi-oxd (approved):
I was able to save an advisory with several relational fields. Looks great 👏

PR #722, opened 2023-05-24
CM-653: Sort advisories by listing rank

molund:
### Jira Ticket:
CM-653

### Jira Ticket URL:
https://bcparksdigital.atlassian.net/browse/CM-653

### Description:
The numeric public-advisory field ‘listingRank’ is used to override the default sort order of advisories on a park page.  Occasionally, an advisory (or advisories) needs to be pinned to the top of the list. A higher number is equivalent to a higher priority on the page.

This was a bit more complicated than expected because there is a bug in the way Strapi sorts nulls on Postgres databases.  As a work-around, I had to ensure that the listingRank is set to zero and is never null.  
https://github.com/strapi/strapi/issues/12593

As part of this change, `loadAdvisories` was also refactored and optimized with the following goals:

1. Reduce code duplication between sites and parks
2. Optimize the JSON payload for public advisories on park pages to eliminate unneeded fields. This included creating a new endpoint called "public-advisories/items" for loading advisories on park pages without loading large text fields like park descriptions.

### DIFF
--- a/src/admin/src/components/page/advisory/Advisory.js
+++ b/src/admin/src/components/page/advisory/Advisory.js
@@ -96,7 +96,7 @@ export default function Advisory({
   const [displayUpdatedDate, setDisplayUpdatedDate] = useState(false);
   const [notes, setNotes] = useState("");
   const [submittedBy, setSubmittedBy] = useState("");
-  const [listingRank, setListingRank] = useState("0");
+  const [listingRank, setListingRank] = useState(0);
   const [ toError, setToError] = useState(false);
   const [toDashboard, setToDashboard] = useState(false);
   const { keycloak, initialized } = useKeycloak();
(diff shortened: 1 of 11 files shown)


src/admin/src/components/page/advisory/Advisory.js
```diff
@@ -96,7 +96,7 @@ export default function Advisory({
   const [displayUpdatedDate, setDisplayUpdatedDate] = useState(false);
   const [notes, setNotes] = useState("");
   const [submittedBy, setSubmittedBy] = useState("");
-  const [listingRank, setListingRank] = useState("");
+  const [listingRank, setListingRank] = useState("0");
```
ayumi-oxd, line 99: If the type of `listingRank` is Number, the initial state will be Number too.
```
const [listingRank, setListingRank] = useState(0);
```

ayumi-oxd:
One nitpicking but other than that it looks great 👍

ayumi-oxd (approved):

PR #795, opened 2023-06-21
CM-637: Add hideStandardCallout to park-activity and park-facility

ayumi-oxd:
### Jira Ticket:
CM-637

### Description:
- Delete `appendStandardCalloutOverride`
- Add `hideStandardCallout`
- Display `appendStandardCalloutText` if `appendStandardCalloutText` is not null
- Hide `appendStandardCalloutText` if `hideStandardCallout` is TRUE

### DIFF
--- a/src/cms/src/api/park-activity/content-types/park-activity/schema.json
+++ b/src/cms/src/api/park-activity/content-types/park-activity/schema.json
@@ -30,7 +30,7 @@
     "isActive": {
       "type": "boolean"
     },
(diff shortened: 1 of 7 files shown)


molund:
Amanda called the boolean filed `hideStandardCallout` in the ticket, not `hideAppendStandardCallout`

I think I agree with her naming since "hide" and "append" are both verbs, and they contradict each other.

molund (approved):

PR #812, opened 2023-07-06
CM-701: Reduced header image height on park pages

foley-lynn:
### Jira Ticket:
CM-701
CM-702

### Description:
Changed parks page header photos to be less tall, smoothed it out responsively. Still not consistent with non-park pages.
Added breadcrumbs to mobile view for non-parks pages. Tried to smooth out spacing a bit, but still very different than park pages.


ayumi-oxd (approved):
LGTM 👏

PR #837, opened 2023-07-14
CM-801: Remove unused fields from public-advisories/items

molund:
### Jira Ticket:
CM-801

### Description:
Removed fields from this API call that are never used on the website.  The lists grow very large when a public advisory is associated with many parks.

### DIFF
--- a/src/cms/src/api/public-advisory/controllers/public-advisory.js
+++ b/src/cms/src/api/public-advisory/controllers/public-advisory.js
@@ -103,28 +103,15 @@ module.exports = createCoreController(
             let pagination;
 
             ctx.query.populate = {
-                accessStatus: { fields: '*' },
-                eventType: { fields: '*' },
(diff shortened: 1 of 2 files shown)


ayumi-oxd:
I had an error when I run `gatsby build`:
```
ERROR #95313 

Building static HTML failed for path "/intro/"

See our docs page for more info on this error:
https://gatsby.dev/debug-html


  19 |
  20 |  // Split the array in 2 parts
> 21 |  var left = components.slice(0, split);
     | ^
  22 |  var right = components.slice(split);
  23 |
  24 |  return Array.prototype.concat.call([], decodeComponents(left),
decodeComponents(right));
```

I think it'll be fixed if you have this commit: https://github.com/bcgov/bcparks.ca/pull/818

ayumi-oxd (approved):
LGTM

PR #880, opened 2023-08-02
CM-846: Change hyphen to "to" when used in a sentence

molund:
### Jira Ticket:
CM-846

### Description:

- Changed hyphen to "to" when used in a sentence on the dates page
- Fixed logic error that was preventing 'year round' from appearing.


manuji (approved):
Looks good.

PR #882, opened 2023-08-03
CM-818: Terraform for AWS OpenSearch

molund:
### Jira Ticket:
CM-818

### Description:
AWS Terraform config for OpenSearch.  

Based on patterns from https://github.com/bcgov/bcparks-ar-admin/tree/main/terraform and https://github.com/bcgov/workbc-jb/blob/master/terraform/dev/es.tf


manuji (approved):
Approved!

PR #883, opened 2023-08-04
CM-841: Reduce breadcrumb padding on small screens to match content

molund:
### Jira Ticket:
CM-841

### Description:
Reduce the breadcrumb padding on small screens to match content, 

The issue was only happening on screens smaller than 415px because the left and right padding was being reduced on the content below at the 415px breakpoint.  I added some extra CSS to also reduce the padding on the breadcrumbs at sizes below 415px.


manuji (approved):
Approved

PR #884, opened 2023-08-04
CM-852: Special text for year-round parks and marine parks

molund:
### Jira Ticket:
CM-852

### Description:
- Show "Limited services" instead of "No services" in the winter text for parks that are open year round
- Remove the word "gate" from marine parks (because they have beaches, not gates)


manuji (approved):
Approved

PR #918, opened 2023-08-21
CM-673: Add expand/collapse all button to accordions

ayumi-oxd:
### Jira Ticket:
CM-673

### Description:
- Add expand/collapse all button to activity/facility/camping/dates of operation section
- Update expand/collapse all button for advisories section

### DIFF
--- a/src/gatsby/src/components/park/advisoryDetails.js
+++ b/src/gatsby/src/components/park/advisoryDetails.js
@@ -2,10 +2,10 @@ import React, { useState } from "react"
 import PropTypes from "prop-types"
 import { parseJSON, format } from "date-fns"
 import Accordion from "react-bootstrap/Accordion"
-import Button from "react-bootstrap/Button"
 import Col from "react-bootstrap/Col"
 import Container from "react-bootstrap/Container"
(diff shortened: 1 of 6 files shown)


molund:
Annotation #2 on the design says "The chevron will behave the same as when accordions are expanded and collapsed **by rotating**". The accordions are animated, so I think these also need to be animated.

molund:
Looks good to me, but there are a couple of niggly things.
1. The "expand all" / "collapse all" for active advisories is designed completely differently (this is more of a question for UX)
2. If I expand all the accordions in a section manually, then it still says "expand all". Likewise, if I click "expand all" and then collapse all the accordions manually, then it still says "collapse all".  But this is consistent with advisories so maybe it's okay.

molund (approved):

molund:
I just noticed that Golden Ears Park has an accordion for Reservations under Camping that appears above the "Expand all" link

molund (changes requested):
I just noticed that Golden Ears Park has an accordion for Reservations under Camping that appears above the "Expand all" link

molund (approved):

PR #922, opened 2023-08-22
CM-858: Fix back button issue on park page

ayumi-oxd:
### Jira Ticket:
CM-858

### Description:
- It works the same as before on my local, but I'm not sure the back button issue will be fixed on the live site

### Todo:
- Testing it on Alpha-dev after merging

### DIFF
--- a/src/gatsby/src/templates/park.js
+++ b/src/gatsby/src/templates/park.js
@@ -163,10 +163,9 @@ export default function ParkTemplate({ data }) {
   }, [apiBaseUrl, park.orcs])
 
   useEffect(() => {
-    if (window.location.hash && !isLoadingProtectedArea && !isLoadingAdvisories) {
-      const hashSave = window.location.hash
-      window.location.hash = ""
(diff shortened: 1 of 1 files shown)


src/gatsby/src/templates/park.js
```diff
@@ -163,10 +163,9 @@ export default function ParkTemplate({ data }) {
   }, [apiBaseUrl, park.orcs])
 
   useEffect(() => {
-    if (window.location.hash && !isLoadingProtectedArea && !isLoadingAdvisories) {
-      const hashSave = window.location.hash
-      window.location.hash = ""
-      window.location.hash = hashSave
+    if (window.location.href && !isLoadingProtectedArea && !isLoadingAdvisories) {
```
molund, line 166: `window.location.href` will always be true.  Are you sure the if condition needs to change?

ayumi-oxd: Good catch, it works without it.

molund (approved):

PR #923, opened 2023-08-23
NOBUG: Fix redirecting issue on park page

ayumi-oxd:
### Jira Ticket:
None

### Description:
- `window.location.replace` should work only if the page URL has a hash


molund:
It works on my box

molund (approved):

PR #925, opened 2023-08-23
CM-792: Refactor filters

molund:
### Jira Ticket:
CM-792

### Description:
- Renamed searchFilter to mobileFilters
- Split desktopFilters into a separate file and reduced duplication with mobileFilters.


ayumi-oxd (approved):
LGTM

PR #951, opened 2023-08-31
CM-906: Combine backcountry camping and wilderness camping

molund:
### Jira Ticket:
CM-906

### Description:
Removed the wilderness camping filter, and converted all wilderness camping facilities to backcountry camping.  

Also fixed a bug where the backcountry camping and vehicle-accessible camping icons were missing.


ayumi-oxd (approved):
LGTM

PR #961, opened 2023-09-06
CM-866: Add css to override material ui style when page is loaded

ayumi-oxd:
### Jira Ticket:
CM-866

### Description:
- Add `!important` to the original CSS to override material ui style when the page is loaded

### DIFF
--- a/src/gatsby/src/styles/staticLanding1.scss
+++ b/src/gatsby/src/styles/staticLanding1.scss
@@ -108,13 +108,13 @@
   }
 }
 
-.parks-search-button {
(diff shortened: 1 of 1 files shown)


molund (approved):

molund:
LGTM.  

Instead of using `!important` you could probably make the selectors more specific.

e.g. use `button.parks-search-button` instead of `.parks-search-button`

The official Gatsby documentation suggests using BEM, but I think that would be overkill
gatsbyjs.com/docs/how-to/styling/global-css/

PR #962, opened 2023-09-06
CM-906: Fix bug combining wilderness camping with backcountry camping

molund:
### Jira Ticket:
CM-906

### Description:
Fixed a coding error that was causing lodash uniqBy to behave incorrectly.  Re-indexing is required after this change.


manuji (approved):
Approved! Thank you

PR #964, opened 2023-09-06
CM-904: Show 6 areas above the fold instead of 5

molund:
### Jira Ticket:
CM-904

### Description:
Show 6 search areas above the fold on desktop instead of 5 (added Thompson based on analytics)

Note: after merging this change the `2023.08.31T00.00.07.search-area.js` migration needs to be run on environments where it has already been applied.  Run this command in Postgres and then restart Strapi. 
```
delete from strapi_migrations where name = '2023.08.31T00.00.07.search-area.js';
```


manuji (approved):
Approved

PR #965, opened 2023-09-06
CM-774: Disable manual scroll restoration on urls with a hash

molund:
### Jira Ticket:
CM-774

### Description:
Disable manual scroll restoration on urls with an anchor hash


manuji (approved):
Approved

PR #983, opened 2023-09-13
CM-808: Format area size number with a thousands separator

ayumi-oxd:
### Jira Ticket:
CM-808

### Description:
-  Format area size number with a thousands separator

### DIFF
--- a/src/gatsby/src/components/park/about.js
+++ b/src/gatsby/src/components/park/about.js
@@ -43,16 +43,16 @@ export default function About({
                   <strong>Size:</strong>
                   {marineProtectedArea === "Y" ?
                     <>
-                      {" "}{park.totalArea.toLocaleString()} hectares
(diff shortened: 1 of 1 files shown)


src/gatsby/src/components/park/about.js
```diff
@@ -43,16 +43,16 @@ export default function About({
                   <strong>Size:</strong>
                   {marineProtectedArea === "Y" ?
                     <>
-                      {" "}{park.totalArea} hectares
+                      {" "}{park.totalArea.toLocaleString()} hectares
```
molund, line 46: I suggest we force the locale to en-CA
`park.totalArea.toLocaleString("en-CA")`
Otherwise, we are at the mercy of the user's browser settings.

molund (changes requested):

molund (approved):

PR #985, opened 2023-09-13
CM-782: Updates to active advisory location display

molund:
### Jira Ticket:
CM-782

### Description:
- Use the same treatment for Regions and Sections on the active advisories page that was already being used for Fire Centres
- Show an extra pill for "Additional parks" if there are parks included in the advisory that are not associated with the displayed Fire Centres, Regions or Sections.

### DIFF
--- a/src/gatsby/src/components/advisories/advisoryCard.js
+++ b/src/gatsby/src/components/advisories/advisoryCard.js
@@ -7,11 +7,67 @@ import HTMLArea from "../HTMLArea"
 
 import "../../styles/advisories/advisoryCard.scss"
 
-const AdvisoryCard = ({ advisory, index }) => {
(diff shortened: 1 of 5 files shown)


src/gatsby/src/components/advisories/advisoryCard.js
```diff
@@ -101,6 +143,31 @@ const AdvisoryCard = ({ advisory, index }) => {
                           {fireCentre.fireCentreName}
                         </Link>
                       ))}
```
ayumi-oxd, line 145: I think we need to display the fireZone badges too

molund: Oh you are right. It's not in the annotations, but there is an example that shows "Arrow fire zone"

ayumi-oxd (approved):
LGTM 👏

PR #1001, opened 2023-09-20
CM-976: Fix missing management document type in management documents list

ayumi-oxd:
### Jira Ticket:
CM-976

### Description:
- Fix missing management document type in the management documents list
- Merge it to main since this issue is happening on PROD


molund:
LGTM. I'm pretty sure I broke this one. 

We don't actually use the data that comes from allStrapiManagementDocument, but I guess we need to load it anyway because we have a typeDef for it.

molund (approved):

PR #1008, opened 2023-09-21
CM-951: Add campfire icon to site page

ayumi-oxd:
### Jira Ticket:
CM-951

### Description:
- Add campfire icon to the site page

### DIFF
--- a/src/gatsby/src/templates/site.js
+++ b/src/gatsby/src/templates/site.js
@@ -141,7 +141,7 @@ export default function SiteTemplate({ data }) {
         setIsLoadingAdvisories(false)
       })
     setIsLoadingProtectedArea(true)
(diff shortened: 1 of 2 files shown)


src/gatsby/src/templates/site.js
```diff
@@ -32,6 +33,22 @@ import Seo from "../components/seo"
 
 import { StyledGrid } from "../utils/constants"
 
+const qs = require('qs')
+
+const loadProtectedArea = (apiBaseUrl, orcsId) => {
```
molund, line 38: You can load protected areas by orcs using the findOne() operation onstead of findMany() like this:

https://cms.bcparks.ca/api/protected-areas/8

The default behaviour was overwritten in the controller code by the previous dev team.

molund (approved):

PR #1034, opened 2023-10-03
CM-1000: Add extra geo-points to Elasticsearch on long line segments

molund:
### Jira Ticket:
CM-1000

### Description:

Add extra geo-points to Elasticsearch to fill on gaps along long line segments.  

Manning Park before

Manning Park after


ayumi-oxd (approved):
LGTM

PR #1036, opened 2023-10-05
CM-799: Add auto suggest to park name search field

ayumi-oxd:
### Jira Ticket:
CM-799

### Description:
- Add auto suggest to park name search field
- Begins immediately as a user starts typing
- Show 8 suggestions on desktop, 4 on mobile 
- Highlight the predictive part of the suggestion (the part that is not what they typed)
- Highlight the active suggestion on hover by underlining it
- When a suggestion is selected, update the text in the search box to match the selection
- Text wrap auto-suggest options into multiple lines when results are too wide for the screen


src/cms/src/api/protected-area/services/search.js
```diff
@@ -304,7 +304,7 @@ module.exports = ({ strapi }) => ({
         {
           multi_match: {
             query: searchText,
-            fuzziness: "AUTO",
+            fuzziness: 1,
             type: "best_fields",
             fields: ["parkNames^2", "protectedAreaName^5"],
             operator: "or"
```
molund, line 310: Let's change this to "and" so we get less bad results

molund (approved):
Let's merge this so Amanda can do a full UX review.  It seems to be working well for me, but I'm still uncertain if we should be using Elasticsearch for the autocomplete or just loading all the parks onto client-side with Gatsby GraphQL and using a custom filterBy. Let's go with this approach for now, but we can revisit it later if there are performance issues.

PR #1069, opened 2023-11-07
CM-1057: autosuggest park name missing in field on park page

molund:
### Jira Ticket:
CM-1057

### Description:
- Fixed missing value in the park name search field when redirecting from the home page
- Fixed bookmarkable URL
- Fixed an issue where the spacebar was triggering a search from the home page, so you couldn't type names with more than 1 word.


ayumi-oxd (approved):
I forgot to have a `value` prop ... 😮 Bookmarkable URL works as expected!

PR #1080, opened 2023-11-16
NOBUG: Fix tabbing issue on find a park page

ayumi-oxd:
### Jira Ticket:
None

### Description:
- Users can select an option with the arrow keys and search parks with the enter key
- User can move to the next element with the tab key
- Reference: https://github.com/ericgio/react-bootstrap-typeahead/issues/461


molund (approved):
It works great

PR #1091, opened 2023-11-23
CM-970: Connect to park names data register

molund:
### Jira Ticket:
CM-970

### Description:
Added a new ETL job to sync park names data with the Park Names Data Register


marklise:
@molund I see your cronjob for the park names is `15 * * * *`, since park names don't change that often, is it possible you could dial down the cron to say once per day when connecting to the data registry?

molund:
> @molund I see your cronjob for the park names is `15 * * * *`, since park names don't change that often, is it possible you could dial down the cron to say once per day when connecting to the data registry?

@marklise There is some new logic in the main cron script so it only ends up running the parknames task once per day:
https://github.com/bcgov/bcparks.ca/blob/2848286e02e3bbe88edbdb9f5f12a18ab9629958/src/etl/index.js
It will only run between 8:00am and 9:00am UTC, and since the script only runs once per hour it will end up running once per day at either 12:15am PST or 1:15am PDT.

marklise:
> > @molund I see your cronjob for the park names is `15 * * * *`, since park names don't change that often, is it possible you could dial down the cron to say once per day when connecting to the data registry?
> 
> @marklise There is some new logic in the main cron script so it only ends up running the parknames task once per day: https://github.com/bcgov/bcparks.ca/blob/2848286e02e3bbe88edbdb9f5f12a18ab9629958/src/etl/index.js It will only run between 8:00am and 9:00am UTC, and since the script only runs once per hour it will end up running once per day at either 12:15am PST or 1:15am PDT.

👍 Thanks for the clarification - didn't dive into the script itself, just saw the cron entry!

ayumi-oxd (approved):

PR #1118, opened 2023-12-05
CM-1125: Fix issue with selected city getting cleared

molund:
### Jira Ticket:
CM-1125

### Description:
Fix issue with selected city getting cleared when loading a bookmarkable URL or searching from the homepage


ayumi-oxd (approved):
Works as expected 👏

PR #1143, opened 2023-12-22
CM-990: Add Park Access Status to Operating Dates page

molund:
### Jira Ticket:
CM-990

### Description:
Add park access status to the park operating dates page.

- This change requires public permissions to be set on a new Strapi endpoint: _"public-advisory.getAccessStatusesByProtectedArea"_
- Layout has been roughly implemented, but needs a design review


ayumi-oxd (approved):
LGTM, the build was successful too

PR #1144, opened 2024-01-02
CM-1158: Change megamenu logic using url

ayumi-oxd:
### Jira Ticket:
CM-1158

### Description:
- Change megamenu logic using url


molund:
Since it's being used 3 times, I think we should declare a constant.

e.g. 
```
const ROOT_MENU_URL = '/';
```

molund (approved):

PR #1153, opened 2024-01-05
CM-1154: Hide dates of operation text if hasParkGate is false

ayumi-oxd:
### Jira Ticket:
CM-1154

### Description:
- Hide dates of operation text if hasParkGate is false

### DIFF
--- a/src/gatsby/src/components/park/parkDates.js
+++ b/src/gatsby/src/components/park/parkDates.js
@@ -160,16 +160,16 @@ export default function ParkDates({ data }) {
 
   // -------- Operating Dates --------
 
-  const thisYear = new Date().getFullYear()  
+  const thisYear = new Date().getFullYear()
 
   // Overall operating dates for parks, to display above subareas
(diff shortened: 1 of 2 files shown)


src/gatsby/src/components/park/parkDates.js
```diff
@@ -260,7 +260,7 @@ export default function ParkDates({ data }) {
                   />
                   {parkStatusText}
                 </h4>
-                {parkDates && (
+                {(parkDates && parkOperation.hasParkGate) && (
```
molund, line 263: hasParkGate is a nullable boolean.  What is the intended behaviour when it's null?

ayumi-oxd: Good question, I'll ask Amanda.

molund: At the standup on January 15th, BC Parks decided to treat null as true.

molund (approved):

PR #1165, opened 2024-01-13
CM-1148: Park access status - update logic to include gates

molund:
### Jira Ticket:
CM-1148

### Description:

- Added a new "Seasonal restrictions" status which is triggered by park gates closures and subarea closures

Testing notes:
- you'll need to re-index elasticsearch after applying this change because `subareaTypeId` was renamed to `subAreaTypeId`.
- parkSubAreaType and parkSubArea must each have at least one record with a value for `closureAffectsAccessStatus`
> - frontcountry camping. day use, day use area, and picnic area sub-area-types should all be set to true, everything else should be `false`
> - set all frount country campgrounds subareas with "overflow" in their name to `false`
> - only parks that would otherwise show as `Open` will show as Seasonal restrictions when the criteria is met
> - closed and inactive subareas are ignored
> - sub area dates must be explicity set to active

### DIFF
--- a/src/elasticmanager/transformers/park/operatingDates.js
+++ b/src/elasticmanager/transformers/park/operatingDates.js
@@ -26,7 +26,7 @@ const convertParkOperationSubAreas = function (parkOperationSubAreas) {
       const sa = {
         isActive: true,
         isOpen: true,
-        subareaTypeId: subarea?.parkSubAreaType?.id,
+        subAreaTypeId: subarea?.parkSubAreaType?.id,
         isIgnored: subarea.closureAffectsAccessStatus === null ? null : !subarea.closureAffectsAccessStatus,
         parkOperationSubAreaDates: subarea.parkOperationSubAreaDates
           .filter(d =>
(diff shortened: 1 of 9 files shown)


ayumi-oxd:
Looks good. My only concern is there might be a Gatsby build error when we merge it into dev. Gatsby build was successful when I added a value for `closureAffectsAccessStatus` in parkSubAreaType and parkSubArea but it didn't work for the first build without any value for `closureAffectsAccessStatus`. 
Adding type to `gatsby-node.js` might help if there's no value on any Strapi environment.

molund:
> Looks good. My only concern is there might be a Gatsby build error when we merge it into dev. Gatsby build was successful when I added a value for `closureAffectsAccessStatus` in parkSubAreaType and parkSubArea but it didn't work for the first build without any value for `closureAffectsAccessStatus`. Adding type to `gatsby-node.js` might help if there's no value on any Strapi environment.

Good idea. I added typedefs

ayumi-oxd (approved):
Works great!

PR #1166, opened 2024-01-15
CM-1151: Seasonal advisory - front-end display

molund:
### Jira Ticket:
CM-1151

### Description:
Added special public advisories to park pages for gate closures.


ayumi-oxd (approved):
LGTM

PR #1170, opened 2024-01-16
CM-1201: Add Park Access Status to Operating Dates page

molund:
### Jira Ticket:
CM-1147
CM-1201

### Description:
- Park access status - update to include ", check advisories" when it is not "Open"
- Add Park Access Status to Operating Dates page


ayumi-oxd (approved):
Looks great!

PR #1174, opened 2024-01-17
CM-1200: Change the last section height on park page to enable scrolling

ayumi-oxd:
### Jira Ticket:
CM-1200

### Description:
- Change the last section height on park page and static content page to enable scrolling
- Remove unnecessary code

### DIFF
--- a/src/gatsby/src/styles/parks.scss
+++ b/src/gatsby/src/styles/parks.scss
@@ -55,6 +55,9 @@
         }
       }
     }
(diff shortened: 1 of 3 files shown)


molund:
It might be better to set it based on the vertical height of the browser minus the height of the footer instead of a fixed size.

e.g. `calc(100vh - 375px)`

molund (approved):

PR #1176, opened 2024-01-18
CM-1201: Update gate logic and display

molund:
### Jira Ticket:
CM-1201
CM-1154

### Description:
Minor updates to gate logic and display based on UX and QA feedback

CM-1201: 
- Reverted back to single-line accessStatus/gateText display on larger devices
- Streamlined logic for determining when to add a period in the text
- Updated some logic that was adding `&nbsp;` characters to date phrases so it is used more sparingly. It was messing up formatting in some places where it wasn't needed.

CM-1154:
- Removed the check for marineProtectedArea
- Don't hide the gate sentence when `hasParkGate === false`. Instead just remove the word "gate" from the sentence.


ayumi-oxd (approved):
LGTM, thanks for updating it!

PR #1184, opened 2024-01-24
CM-1222: Fix bug in seasonal access status for subareas

molund:
### Jira Ticket:
CM-1222
CM-1151

### Description:
The wrong keyword was used to exit the for loop. 
Also updated a copy error.


molund:
> I couldn't see CM-1122 on Jira, did you mean ticket number CM-1222 plus CM-1151? The fix itself looks good to me 🙂

You're right, I typed the wrong ticket number. Thanks for catching that.

ayumi-oxd (approved):

PR #1186, opened 2024-01-24
CM-1203: Fix typo in the park search options

ayumi-oxd:
### Jira Ticket:
CM-1203

### Description:
- Fix typo in the park search options


molund:
Could something like this work, using a regex, so you wouldn't have to convert to string to lowercase in the first place?
https://stackoverflow.com/a/47676769

molund (approved):

PR #1195, opened 2024-02-01
CM-997: Change the advisory status from camel case to sentence case

ayumi-oxd:
### Jira Ticket:
CM-997

### Description:
- Change the advisory status from camel case to sentence case

### DIFF
--- a/src/admin/src/utils/AppUtil.js
+++ b/src/admin/src/utils/AppUtil.js
@@ -30,14 +30,8 @@ export function dateCompare(a, b) {
 }
 
 export function camelCaseToSentenceCase(input) {
-  // split the string into words based on capital letters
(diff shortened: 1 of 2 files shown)


src/admin/src/utils/AppUtil.js
```diff
@@ -28,3 +28,16 @@ export function dateCompare(a, b) {
   }
   return 0;
 }
+
+export function camelCaseToSentenceCase(input) {
```
molund, line 32: I think this might be way more complicated than it needs to be.

Couldn't you just do something simple like this?
```
function sentenceCase(input) {
    if (!input) {
        return "";
    }
    return input[0].toUpperCase() + input.substring(1).toLowerCase();
}
```

molund (approved):

PR #1197, opened 2024-02-01
CM-1043: Update colours to align with colours in design library

ayumi-oxd:
### Jira Ticket:
CM-1043

### Description:
- Update colours to align with colours in the design library
- Remove unused CSS classes

### DIFF
--- a/src/gatsby/src/styles/search.scss
+++ b/src/gatsby/src/styles/search.scss
@@ -314,15 +314,26 @@ h3.subtitle {
   }
 
   .park-filter-chip {
-    margin: 0 4px 8px 0 !important;
-    min-width: 130px;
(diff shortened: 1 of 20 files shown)


src/gatsby/src/styles/search.scss
```diff
@@ -1160,7 +1068,7 @@ div[class$="-singleValue"] {
   }
 
   .close-icon:hover {
-    color: $colorGrey !important;
+    color: $colorBackgroundGrey !important;
```
molund, line 1071: Did you mean to use a background color variable for a non-background element?

ayumi-oxd: Good catch! Actually, this class was not needed so I removed it.

PR #1207, opened 2024-02-07
CM-1231: Fix link type was missing when user edits an advisory

ayumi-oxd:
### Jira Ticket:
CM-1231

### Description:
- Fix the link type that was missing when a user edited an advisory


molund (approved):
Looks good to me

PR #1218, opened 2024-02-13
CM-1278: Prevent mousewheel in input of type number

ayumi-oxd:
### Jira Ticket:
CM-1278

### Description:
- Prevent mousewheel in input of type number


molund:
Neat. I've never seen anyone use onWheel before

molund (approved):

PR #1220, opened 2024-02-13
CM-1285: Revamp public advisory audit-tracking fields

molund:
### Jira Ticket:
CM-1285

### Description:

1. The `createdBy` field (which should hold a person's full name) hasn't been getting saved to Strapi since the Strapi 4 update.  This is happening because it conflicts with the built-in Strapi database field `created_by_id`.  
      (a.) A new field called `createdByName `will be added to store the missing value
      (b.) `createdByName` should only exist in the publicAdvisoryAudit collection and not in publicAdvisory because https://cms.bcparks.ca/api/public-advisories is publicly accessible, and staff names should not be appearing in the public API unnecessarily.

2. The `modifiedBy` field was in both publicAdvisoryAudit and publicAdvisory. It will be removed from publicAdvisory for the reason stated in 1b above.

4. Three new fields `createdByRole`, `modifiedByRole` and `isUrgentAfterHours` will be added to publicAdvisoryAudit to help to determine when email notifications should be triggered for CM-1259

### DIFF
--- a/src/admin/src/components/page/advisory/Advisory.js
+++ b/src/admin/src/components/page/advisory/Advisory.js
@@ -790,7 +790,6 @@ export default function Advisory({
           note: notes,
           submittedBy: submittedBy ? submittedBy : submitter,
           createdDate: moment().toISOString(),
-          createdBy: keycloak.tokenParsed.name,
           advisoryDate: advisoryDate,
           effectiveDate: startDate,
           endDate: endDate,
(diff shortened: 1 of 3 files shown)


ayumi-oxd (approved):
LGTM - As a submitter, I confirmed my name in the `createdBy` field and `submittedBy` field, and my role in the `createdByRole` field when I created an advisory. And, `isUrgentAfterHours` is automatically applied. Also, I confirmed my role in the `modifiedByRole` field when I edited an advisory.

PR #1231, opened 2024-02-20
CM-1250: Moved the Strapi cron logic into a service & controller

molund:
### Jira Ticket:
CM-1250

### Description:
- This is the first in a series of PR's for CM-1250
- Moved the logic from cron-task.js into an API service and controller action that can be externally triggered  
    - This controller action will be called by the service (formerly known as the Elasticmanager). 
    - It was easier to do it this way than to use the REST API directly because the existing code heavily used the Strapi entity service.  
- Refactored the existing public-advisory service to split out search functionality and cron functionality, similar to how the protected-area API is architected.


ayumi-oxd (approved):
Works as expected.

PR #1246, opened 2024-02-27
CM-1169: Email link to advisories

molund:
### Jira Ticket:
CM-1169

### Description:
Added an alternate path for linking to advisories in the staff portal using the advisoryNumber instead of the publicAdvisoryAuditId


ayumi-oxd (approved):
LGTM, the redirects worked when I change `ADMIN_URL=http://127.0.0.1:3000` to `ADMIN_URL=http://localhost:3000`

PR #1250, opened 2024-02-29
CM-480: Add footer menu collection to strapi

ayumi-oxd:
### Jira Ticket:
CM-480

### Description:
- Add footer menu collection to Strapi
- Remove footer test files
- Add campfire ban test files as a test sample file
- Add footer menu collection data for migration
- Remove isExternalUrl field
- Add a condition not to display more than 3 sections if users add new ones


src/cms/database/migrations/2024.03.01T00.00.10.footer-menu.js
```diff
@@ -0,0 +1,142 @@
...
+    "parent_id": 3,
+  }
+]
+
+async function up(knex) {
+
+  let i, j, rowsToInsert, chunk = 200;
+
+  // insert rows in 200 row batches
+  if (await knex.schema.hasTable('footer_menus')) {
```
molund, line 132: We'll also need to create a "post deployment tasks" ticket for this release with a step to delete the migration from Postgres and restart one of the CMS pods.  

The table won't get created until AFTER the migration is run, so hasTable will return "false" the first time it runs

ayumi-oxd: https://bcparksdigital.atlassian.net/browse/CM-1323

molund (approved):

PR #1266, opened 2024-03-15
CM-1223: Add natural resource district data

molund:
### Jira Ticket:
CM-1223

### Description:
This PR adds the Natural-resource-district collection and imports the data.  I had to use two migration files to split up the database transaction in order to read the Natural-resource-district data back from the database and get the generated ids. 

After the schema is updated, you'll need to delete two records from the `strapi_migrations` table and then restart Strapi so the migrations run a second time.
```
delete from strapi_migrations where name like '%natural-resource%';
```


ayumi-oxd (approved):
LGTM! I was able to migrate the data to the new Strapi collection, and the Gatsby build was successful.

PR #1269, opened 2024-03-18
CM-1107: Update BC sans font

ayumi-oxd:
### Jira Ticket:
CM-1107

### Description:
- Update BC sans font to v2: https://github.com/bcgov/bc-sans

### DIFF
--- a/src/gatsby/gatsby-browser.js
+++ b/src/gatsby/gatsby-browser.js
@@ -1,4 +1,5 @@
 import "@bcgov/bootstrap-theme/dist/css/bootstrap-theme.min.css"
+import "@bcgov/bc-sans/css/BC_Sans.css"
 import "jquery/dist/jquery.slim"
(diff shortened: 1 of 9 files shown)


src/gatsby/gatsby-browser.js
```diff
@@ -1,4 +1,5 @@
 import "@bcgov/bootstrap-theme/dist/css/bootstrap-theme.min.css"
+import "@bcgov/bc-sans/css/BC_Sans.css"
```
molund, line 2: How was the font getting imported previously? Was it coming from the bootstrap theme?

ayumi-oxd: Yes, I assume it was from the bootstrap: https://github.com/bcgov/bootstrap-theme/tree/master
However, it hasn't been updated for a while so I imported the font file.

PR #1284, opened 2024-04-15
CM-1346: Add snowplow browser tracker

ayumi-oxd:
### Jira Ticket:
CM-1346

### Description:
- Update megamenu link with Gatsby link  
- Add tracking search filter click event

### DIFF
--- a/src/gatsby/gatsby-browser.js
+++ b/src/gatsby/gatsby-browser.js
@@ -1,4 +1,4 @@
-import { newTracker, enableActivityTracking, trackPageView } from '@snowplow/browser-tracker';
+import { newTracker, enableActivityTracking, refreshLinkClickTracking, trackPageView } from '@snowplow/browser-tracker';
 import "@bcgov/bootstrap-theme/dist/css/bootstrap-theme.min.css"
 import "@bcgov/bc-sans/css/BC_Sans.css"
 import "jquery/dist/jquery.slim"
@@ -19,6 +19,8 @@ export const onInitialClientRender = () => {
     minimumVisitLength: 30,
     heartbeatDelay: 10
   });
+
+  refreshLinkClickTracking();
 };
 
 export const onRouteUpdate = ({ location, prevLocation }) => {
--- a/src/gatsby/gatsby-browser.js
+++ b/src/gatsby/gatsby-browser.js
@@ -1,4 +1,5 @@
-import { newTracker, enableActivityTracking, refreshLinkClickTracking, trackPageView } from '@snowplow/browser-tracker';
+import { newTracker, enableActivityTracking, trackPageView } from '@snowplow/browser-tracker';
+import { refreshLinkClickTracking } from '@snowplow/browser-plugin-link-click-tracking';
 import "@bcgov/bootstrap-theme/dist/css/bootstrap-theme.min.css"
 import "@bcgov/bc-sans/css/BC_Sans.css"
 import "jquery/dist/jquery.slim"
--- a/src/gatsby/gatsby-browser.js
+++ b/src/gatsby/gatsby-browser.js
@@ -1,5 +1,3 @@
-import { newTracker, enableActivityTracking, trackPageView } from '@snowplow/browser-tracker';
-import { refreshLinkClickTracking } from '@snowplow/browser-plugin-link-click-tracking';
 import "@bcgov/bootstrap-theme/dist/css/bootstrap-theme.min.css"
 import "@bcgov/bc-sans/css/BC_Sans.css"
 import "jquery/dist/jquery.slim"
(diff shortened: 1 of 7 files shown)


src/gatsby/gatsby-browser.js
```diff
@@ -1,11 +1,28 @@
+import { newTracker, enableActivityTracking, trackPageView } from '@snowplow/browser-tracker';
 import "@bcgov/bootstrap-theme/dist/css/bootstrap-theme.min.css"
 import "@bcgov/bc-sans/css/BC_Sans.css"
 import "jquery/dist/jquery.slim"
 import "bootstrap/dist/js/bootstrap.bundle"
 import "@fortawesome/fontawesome-free/css/all.min.css"
 import "./src/styles/style.scss"
 
+// Snowplow tracking
```
molund, line 9: Is it possible to use the snowplow instance that's already loaded instead of loading another one?
e.g. https://bcparks.ca/_scripts/snowplow.js is already loaded in `bcparks.ca/src/gatsby/gatsby-ssr.js`
```
const HeadComponents = [
    process.env.GATSBY_ENABLE_SNOWPLOW === "true" &&
    <script src="/_scripts/snowplow.js" key="snowplow-js" />,
    ...
];
```

molund: For workbc it looks like we used the windows object for custom snowplow tracking.  
https://github.com/search?q=repo%3Abcgov%2Fworkbc-jb%20snowplow&type=code
workbc was typescript, so we had to cast `window as any`, but that won't be needed in js.

molund: I think calls to `refreshLinkClickTracking` like on workBC will probably also be needed in most cases.
https://docs.snowplow.io/docs/collecting-data/collecting-from-own-applications/javascript-trackers/web-tracker/tracking-events/link-click/

ayumi-oxd: Seems these two are a bit different. The `gatsby-ssr.js` is used during the server-side rendering process. This script tag includes the Snowplow JavaScript file (snowplow.js) in the pages. This is necessary for the Snowplow tracking to work. The `gatsby-browser.js` is used for customizing and extending default settings affecting the browser. This is where you initialize the Snowplow tracker and set up event tracking. This code runs in the browser after the page has loaded. But I'm not 100% sure tbh, I couldn't find many examples of Gatsby and Snowplow 🥲

src/gatsby/src/components/search/filter.js
```diff
@@ -59,13 +58,13 @@ const Filter = ({ filterItems, selectedFilterItems, handleFilterCheck, filterTyp
...
-                  action: event.target.checked ? 'Check' : 'Uncheck',
-                  label: item.label,
-                  property: 'filter-checkbox',
-                  value: event.target.checked ? 1 : 0
-                })
+                window.snowplow(
+                  "trackStructEvent",
+                  "Checkbox",
+                  event.target.checked ? "Check" : "Uncheck",
+                  item.label, "filter-checkbox",
```
molund, line 65: In addition to the label, it would be good to track the category of the checkbox. e.g. "Popular" / "Area" / Camping / "Activities" / "Facilities"

src/gatsby/src/components/search/filter.js
```diff
@@ -58,10 +58,21 @@ const Filter = ({ filterItems, selectedFilterItems, handleFilterCheck, filterTyp
               checked={checked}
               onChange={event => {
                 handleFilterCheck(item, event)
+                if (typeof window.snowplow === 'function') {
+                  window.snowplow(
+                    "trackStructEvent",
+                    "Checkbox",
```
molund, line 64: The category could be something like 'park-search` or `find-a-park`.  It describes what part of the site you are measuring.

src/gatsby/src/components/search/filter.js
```diff
@@ -58,10 +58,21 @@ const Filter = ({ filterItems, selectedFilterItems, handleFilterCheck, filterTyp
...
+                    event.target.checked ? "Check" : "Uncheck",
```
molund, line 65: The action should capture what UI element was interacted with and what the user did. 
e.g. `filter-select`, `filter-unselect`

molund, line 65: `select-filter` / `unselect-filter`

src/gatsby/src/components/search/filter.js
```diff
@@ -58,10 +58,21 @@ const Filter = ({ filterItems, selectedFilterItems, handleFilterCheck, filterTyp
...
+                    `${item.filterType} - ${item.label}`,
```
molund, line 66: This is good, but you could use the filterType as a property and just pass the label here.

src/gatsby/src/components/search/filter.js
```diff
@@ -58,10 +58,21 @@ const Filter = ({ filterItems, selectedFilterItems, handleFilterCheck, filterTyp
...
+                    "filter-checkbox",
```
molund, line 67: I would change this to `filter-type` and use it to pass `item.filterType` as a value. This allows grouping by the label alone without the group.

src/gatsby/src/components/search/filter.js
```diff
@@ -58,10 +58,21 @@ const Filter = ({ filterItems, selectedFilterItems, handleFilterCheck, filterTyp
...
+                    event.target.checked ? 1 : 0
```
molund, line 68: `item.filterType`

src/gatsby/gatsby-browser.js
```diff
@@ -5,7 +5,32 @@ import "bootstrap/dist/js/bootstrap.bundle"
...
+      }
+    });
+
+    // ping every 30 seconds after 30 seconds
+    window.snowplow("enableActivityTracking", 30, 30);
+    window.snowplow("enableLinkClickTracking");
+    window.snowplow("trackPageView");
+  }
+};
+
```
molund, line 29: I think this was being done in gatsby-ssr so it could be controlled by a server-side environment variable which is set here: https://github.com/bcgov/bcparks.ca/blob/main/.github/workflows/publish-gatsby.yaml

I'll see if I can find a way to control this on our local dev environments without loading it on client render

molund:
Two pageviews are being tracked when the site first loads.  
We will need to update the configmaps in openshift prod & test to remove `window.snowplow('trackPageView');`l.  This should be added to the **post deployment tasks** ticket.  
https://console.apps.silver.devops.gov.bc.ca/k8s/ns/61d198-prod/configmaps/snowplow
https://console.apps.silver.devops.gov.bc.ca/k8s/ns/61d198-test/configmaps/snowplow

molund (approved):
This seems to be working.  It cannot be tested on alpha-dev or dev, but it should work on alpha-test and test.   Please see my note about post deployment tasks

ayumi-oxd:
@molund Sounds good. Let's test it on Alpha-TEST.

molund:
I have removed the call to `window.snowplow('trackPageView');` from test/alpha-test
https://console.apps.silver.devops.gov.bc.ca/k8s/ns/61d198-test/configmaps/snowplow

PR #1285, opened 2024-04-16
CM-1283: Email notification when a name change occurs

molund:
### Jira Ticket:
CM-1283

### Description:
Send an email notification when the parknames ETL task changes a park name.


ayumi-oxd (approved):
Works as expected!

PR #1297, opened 2024-04-30
CM-1014: Add dropdown for status and urgency to filter public advisories

ayumi-oxd:
### Jira Ticket:
CM-1014

### Description:
- Add dropdown for status and urgency to filter public advisories
- Change the filter type to "date" for the dates columns


molund (approved):
Looks good

PR #1318, opened 2024-05-22
CMS-191: Display event options without typing

ayumi-oxd:
### Jira Ticket:
CM-191

### Description:
- Display event options without typing on the Active Advisories page


molund:
I changed your PR to merge to alpha instead of develop

PR #1319, opened 2024-05-22
CMS-62: Add protectedArea and site relation to parkPhoto

ayumi-oxd:
### Jira Ticket:
CMS-62

### Description:
- Add protectedArea and site relation to parkPhoto
- If protectedArea/site is added to parkPhoto, its orcs/orcsSiteNumber will be updated too
- No change on FE at this moment
- Data migration will be needed
- Nearby park section file should be updated

### DIFF
--- a/src/cms/src/api/park-photo/content-types/park-photo/lifecycles.js
+++ b/src/cms/src/api/park-photo/content-types/park-photo/lifecycles.js
@@ -22,6 +22,21 @@ const getOrcs = async function (event) {
   return photo?.orcs;
 };
 
+const getOrcsSiteNumber = async function (event) {
+  let { where } = event.params;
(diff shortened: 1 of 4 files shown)


molund (approved):
This looks good, but I think we should disable the "orcs" and "orcsSiteNumber" fields on the Strapi form when we deploy it.  We can do this by going into "Configure the view", editing the fields, and setting "Editable field" to "FALSE".

ayumi-oxd:
> This looks good, but I think we should disable the "orcs" and "orcsSiteNumber" fields on the Strapi form when we deploy it. We can do this by going into "Configure the view", editing the fields, and setting "Editable field" to "FALSE".

Sounds good, I'll add this note to the post-deployment task.

PR #1323, opened 2024-05-28
CMS-26: Update tldr section

ayumi-oxd:
### Jira Ticket:
CMS-25
CMS-26
CMS-31
CMS-221

### Description:
- Update tldr section
- Add area map images with/without label
- Update access status icons and advisory status icons
- Update maps and location section


molund (approved):
Looks good.  I noticed a few UX & design issues, but everything seems to work from a dev perspective.

PR #1331, opened 2024-05-31
CMS-220: Migration for CMS-62

molund:
### Jira Ticket:
CMS-220

### Description:
This migration will link the park photos to sites and protected areas based on existing keys


ayumi-oxd (approved):
Works as expected!

PR #1336, opened 2024-06-05
CMS-211: Regenerate Kong config

molund:
### Jira Ticket:
CMS-211

### Description:
This fixes an error reported by BC Parks where https://bcparks.api.gov.bc.ca/api/park-operation-dates was not accessible via the API gateway.  

In addition, two new collection types were added to the API gateway
- emergency-alerts
- natural-resource-districts

The configuration of the documentation plugin (which is used to generate the Kong config) was also updated to hide collection types that are only intended to be used internally for the operation of the website.  

- public-advisory-audit
- queued-task
- search-indexing
- statutory-holiday
- geo-shape
- search-city
- footer-menu
- menu
- website
- uploads plugin
- users-permissions plugin


ayumi-oxd (approved):
LGTM. I was able to access https://cms.bcparks.ca/api/park-operation-dates and https://bcparks.api.gov.bc.ca/api/park-operation-dates. Also, I confirmed the kong documentation is updated on my local.

PR #1339, opened 2024-06-07
CMS-225: More encompassing null checks for accessStatus

molund:
### Jira Ticket:
CMS-123

### Description:
Wrapped the null check around the entire code block


ayumi-oxd (approved):
LGTM

PR #1362, opened 2024-06-25
CMS-203: Reorganize park page and site page

ayumi-oxd:
### Jira Ticket:
CMS-203
CMS-236

### Description:
- Reorganize park page and site page, change the order of sections
- Add the contact section
- Add old anchor links to the section subheading as a fallback


src/gatsby/src/components/park/natureAndCulture.js
```diff
@@ -2,11 +2,67 @@ import React from "react"
```
molund, line 2: Should this file be renamed, since this isn't the natureAndCulture section anymore?

molund (approved):
Looks good, aside from my comments about updating file names to match the new titles.  

I couldn't get this to run on my local, but I think it's something to do with the Strapi 4.25 upgrade, not with your code.

PR #1365, opened 2024-06-26
CMS-29: Add visitor guidelines

ayumi-oxd:
### Jira Ticket:
CMS-29

### Description:
- Add a visitor guidelines section to the Park page
- Add `guidelineType` and `parkGuideline` collections
- Add a relation to `protectedArea` and `parkGuideline`


src/cms/src/api/guideline-type/content-types/guideline-type/schema.json
```diff
@@ -0,0 +1,49 @@
...
+    "singularName": "guideline-type",
+    "pluralName": "guideline-types",
+    "displayName": "Guideline-type",
+    "description": ""
+  },
+  "options": {
+    "draftAndPublish": true
+  },
+  "attributes": {
+    "guidelineNumber": {
```
molund, line 14: What does the guidelineNumber do?

src/cms/src/api/guideline-type/content-types/guideline-type/schema.json
```diff
@@ -0,0 +1,49 @@
...
+      "required": true
+    },
+    "defaultDescription": {
+      "type": "customField",
+      "options": {
+        "preset": "toolbar"
+      },
+      "customField": "plugin::ckeditor5.CKEditor",
+      "required": true
+    }
```
molund, line 47: Can we include a "defaultOrder" too?  If the order on the park-guideline isn't set then it would use the defaultOrder.  

We use the names "rank" and "precedence" on some other sorting fields.  But I'm okay with introducing a new field name in this case because I can never tell from these other labels if they are supposed to be ascending or descending.

src/cms/src/api/guideline-type/content-types/guideline-type/schema.json
ayumi-oxd: Removed `guidelineNumber`

src/cms/src/api/guideline-type/content-types/guideline-type/schema.json
ayumi-oxd: Change `order` to `rank`, also added `defaultRank` to `guidelineType`

src/gatsby/src/components/park/visitorGuidelines.js
```diff
@@ -0,0 +1,59 @@
...
+    </Row>
+  )
+}
+
+export default function VisitorGuidelines({ guidelines }) {
+  // Filter isActive and sort by order
+  const sortedGuidelines =
+    guidelines.filter(guide => guide.isActive).sort((a, b) => {
+      // Check if both have rank to override defaultRank
+      if (a.rank && b.rank) {
```
molund, line 45: I don't think the fallback will work properly in a case where **a** has its own rank and **b** needs to use the defaultRank

molund (approved):

PR #1366, opened 2024-06-28
CMS-163: Fix error running strapi 4.25 in non-dev mode

molund:
### Jira Ticket:
CMS-163

### Description:
This block of code never worked in non-dev mode anyway, but now it's causing Strapi admin to fail completely with the error  "process is not defined". The error is on both alpha-dev and alpha-test.
@ayumi-oxd should we just remove this code completely, or keep it because it did work on localdev? I'm leaning toward deleting the entire switch statement

Update: the code has been removed completely


ayumi-oxd (approved):
I'm OK with removing it. Thanks for letting me know!

PR #1388, opened 2024-07-26
CMS-112: Update accordion function

ayumi-oxd:
### Jira Ticket:
CMS-112

### Description:
-  If there is only one accordion in a section, have it default to expanded and hide the Expand all and expand/collapse functionality


molund (approved):
Once this is merged, I will need to rebase my camping section changes onto it

PR #1394, opened 2024-07-30
CMS-275: Add conditions to display reservations required

ayumi-oxd:
### Jira Ticket:
CMS-275

### Description:
- Add conditions to display reservations required


molund:
I think CMS-246 (which I merged yesterday) might change some of this logic, but maybe not until the second release of the park page redesign.

molund (approved):

PR #1409, opened 2024-08-08
CMS-257: Add fontawesome svg files

ayumi-oxd:
### Jira Ticket:
CMS-257

### Description:
- Remove fontawesome js files
- Add fontawesome svg files


src/gatsby/src/styles/parks.scss
```diff
@@ -481,9 +481,8 @@ h2.section-heading {
   margin-bottom: 16px;
   &--left {
     svg {
-      width: 36px;
-      height: 36px;
-      color: $colorBlue;
+      width: 48px;
+      height: 48px;
```
duncan-oxd, line 485: If you ever need to show these icons in different colors, you can use the currentColor CSS property to behave like the web font and inherit the color of the `span` tag around it 👍

duncan-oxd (approved):
Looks good! I'd love to hear if the size to download the app has changed now that it's only including a few icons

src/gatsby/src/components/fontAwesome.js
```diff
@@ -0,0 +1,66 @@
...
+    "x-twitter": xTwitter
+  }
+  const svg = icons[icon] || icon
+
+  return (
+    <span dangerouslySetInnerHTML={{ __html: svg }} className={className} />
+  )
+}
+
+export default FontAwesome
```
molund, line 66: This approach should work fine, but I'm curious why you didn't follow the same pattern as we have here? 
https://github.com/bcgov/bcparks.ca/blob/main/src/gatsby/src/components/park/staticIcon.js

Sometimes Rob likes to commit new icons directly to GitHub, and the existing pattern allows him to do that without requiring us to modify code.

ayumi-oxd:
> Looks good! I'd love to hear if the size to download the app has changed now that it's only including a few icons

I noticed the size was a bit off after I downloaded it but I have no idea why ... 🙃
Screenshot 2024-08-08 at 11 47 43 AM
Screenshot 2024-08-08 at 11 47 15 AM

molund (approved):

molund (approved):

PR #1413, opened 2024-08-09
NOBUG: Update campground icon

ayumi-oxd:
### Jira Ticket:
None

### Description:
- Update campground icon


duncan-oxd (approved):
Looks good to me. If we need to change the colors of the icons, we can look into doing that with CSS classes too 👍

PR #1427, opened 2024-08-14
CMS-311: Display expand all button if there's one camping accordion including reservations

ayumi-oxd:
### Jira Ticket:
CMS-311

### Description:
- Display expand all button if there's one camping accordion including reservations


duncan-oxd (approved):
Looks good, as long as `activeCampaigns` will never be empty(?)

PR #1428, opened 2024-08-14
CMS-267: Add console debugging for thumbnail images

duncan-oxd:
### Jira Ticket:
CMS-267

### Description:
This adds a bunch of debugging stuff to the console to try to find the cause of the broken images on the alpha-dev and alpha-test servers. 

1. It will print the original URL for all the thumbnails
2. It will print the image URL and the error message in `handleImgError`

We can revert this commit after we solve the issue!


ayumi-oxd (approved):
LGTM

PR #1437, opened 2024-08-19
CMS-279: Add reservation buttons in camping section

ayumi-oxd:
### Jira Ticket:
CMS-279
CMS-370

### Description:
- Add reservation buttons in camping/facilities section
- Move dates of operation from camping accordion

### DIFF
--- a/src/gatsby/src/components/park/parkDates.js
+++ b/src/gatsby/src/components/park/parkDates.js
@@ -8,6 +8,84 @@ import { faChevronUp, faChevronDown } from "@fortawesome/free-solid-svg-icons"
 import HtmlContent from "./htmlContent"
 import SubArea from "./subArea"
 
+export const ReservationButtons = ({ campingTypeCode, parkOperation }) => {
+  const reservationUrlRules = {
+    "backcountry-camping": {
+      buttons: [
+        {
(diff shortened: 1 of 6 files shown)


src/gatsby/src/components/park/parkDates.js
```diff
@@ -8,6 +8,88 @@ import { faChevronUp, faChevronDown } from "@fortawesome/free-solid-svg-icons"
...
+        {
+          label: "Reserve canoe circuit",
+          fieldName: "canoeCircuitReservationUrl"
+        },
+        {
+          label: "Book groupsite",
+          fieldName: "backcountryGroupReservationUrl"
+        },
+        {
+          label: "Book shelter",
```
molund, line 32: I think backcountry shelter goes under the facility-type shelters
see https://alpha-dev.bcparks.ca/garibaldi-park/#facilities

ayumi-oxd: On PROD, "Elfin Lakes shelter" belongs to the "Cabins and huts". Could shelters be either facility or camping depending on its park?

molund (approved):
Looks good.  If we need to move buttons around as a result conversations on MS Teams then we can open a new PR for that.

PR #1442, opened 2024-08-22
CMS-310: Add gradient to expandable description

ayumi-oxd:
### Jira Ticket:
CMS-310

### Description:
- Add gradient to expandable description (Hightlight section and Camping section)


duncan-oxd (approved):
Looks good! That's a nice effect. I put in a small suggestion for the CSS if you want to update it, but it looks nice the way it is.

You could also consider using an `:after` pseudo-element, or putting the "Show more..." button inside the gradient element so you won't need an empty div just for the visual effect.

molund (approved):

duncan-oxd (approved):

PR #1443, opened 2024-08-22
CMS-353: Data migration for nature and culture

molund:
### Jira Ticket:
CMS-353

### Description:

- This script migrates the nature and culture field into 4 new fields
- Only parks with a parsable nature and culture section will be migrated
- To run the migration script, you'll need to delete a strapi migration and restart strapi
`2024.08.22T00.00.20.nature-and-culture.js`
- The line of code that removes the old data from the natureAndCulture field has been commented out in the migration so QA can compare the data

### DIFF
--- a/src/cms/database/migrations/2024.08.22T00.00.20.nature-and-culture.js
+++ b/src/cms/database/migrations/2024.08.22T00.00.20.nature-and-culture.js
@@ -1,0 +1,108 @@
+'use strict'
+
+const cheerio = require('cheerio');
+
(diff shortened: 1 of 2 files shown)


ayumi-oxd (approved):
Works as expected!

manuji:
Hi @molund , reminder to remove the comment on the line of code that removes the old data from the natureAndCulture field for PROD deployment. I think it wasn't removed for TEST environment as well, but that's okay.

PR #1446, opened 2024-08-26
CMS-310: Hide show more button if hr is at the bottom

ayumi-oxd:
### Jira Ticket:
CMS-310

### Description:
- Hide show more button and gradient if hr is at the bottom
- If the height is 261-299 pixels then we don't show the link, but if the height is more than 300 pixels then we show the link 
- We only show 260 pixels of the content by default


molund (approved):
LGTM.  Thanks

PR #1450, opened 2024-08-28
CMS-354: Not display conservation in accordion if the park is ecological reserve

ayumi-oxd:
### Jira Ticket:
CMS-354

### Description:
- Not display conservation in accordion if the park type is ecological reserve


molund (approved):
LGTM.  I confirmed with UX that we never need to show the other 3 fields for eco reserves.

PR #1461, opened 2024-09-06
CMS-388: Display 'first come, first serve' if subarea doesn't have reservation dates

ayumi-oxd:
### Jira Ticket:
CMS-388

### Description:
- Display 'first come, first serve' if subarea doesn't have reservation dates on the park page
- ^ Same as the park operating dates page


ayumi-oxd:
We've decided to leave the current implementation as is.

PR #1463, opened 2024-09-06
CMS-279: Update condition for reservation deep link buttons

ayumi-oxd:
### Jira Ticket:
CMS-279

### Description:
- Update condition for reservation deep link buttons


molund (approved):
Looks good. I confirmed that the only backcountry group site is Croteau Lake Backcountry and it is categories as a group site (not backcountry) in Strapi.  
https://bcparks.ca/strathcona-park/#park-dates-container

PR #1466, opened 2024-09-09
CMS-260: Update condition for the default contact information

ayumi-oxd:
### Jira Ticket:
CMS-260

### Description:
- Update the condition for the default contact information
- Some code refactors

### DIFF
--- a/src/gatsby/src/components/park/contact.js
+++ b/src/gatsby/src/components/park/contact.js
@@ -81,7 +81,6 @@ export default function Contact({ contact, parkContacts, operations }) {
     operations.hasPicnicShelterReservations ||
     operations.hasFrontcountryReservations ||
     operations.hasFrontcountryCabinReservations ||
-    operations.hasBackcountryPermits ||
     operations.hasBackcountryReservations ||
     operations.hasBackcountryGroupReservations ||
     operations.hasBackcountryShelterReservations ||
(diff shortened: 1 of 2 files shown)


src/gatsby/src/components/park/contact.js
```diff
@@ -81,7 +81,6 @@ export default function Contact({ contact, parkContacts, operations }) {
     operations.hasPicnicShelterReservations ||
```
molund, line 81: I think this is actually called hasGroupPicnicReservations

src/gatsby/src/components/park/contact.js
```diff
@@ -81,7 +81,6 @@ export default function Contact({ contact, parkContacts, operations }) {
...
     operations.hasFrontcountryReservations ||
     operations.hasFrontcountryCabinReservations ||
-    operations.hasBackcountryPermits ||
     operations.hasBackcountryReservations ||
     operations.hasBackcountryGroupReservations ||
     operations.hasBackcountryShelterReservations ||
```
molund, line 86: We should also include hasFrontcountryGroupReservations (in place of hasGroupReservations which is noted in the ticket)

src/gatsby/src/components/park/contact.js
```diff
@@ -81,7 +81,6 @@ export default function Contact({ contact, parkContacts, operations }) {
     operations.hasPicnicShelterReservations ||
     operations.hasFrontcountryReservations ||
     operations.hasFrontcountryCabinReservations ||
-    operations.hasBackcountryPermits ||
```
molund, line 84: We should confirm with Amanda that this was actually a bug and not an error in the requirements.

ayumi-oxd: Good catch! Confirmed with Amanda and Brett: https://bcparksdigital.atlassian.net/browse/CMS-260

molund (approved):

PR #1469, opened 2024-09-11
CMS-279: Update reservation button condition

ayumi-oxd:
### Jira Ticket:
CMS-279

### Description:
- Update reservation button condition

### DIFF
--- a/src/gatsby/src/components/park/parkDates.js
+++ b/src/gatsby/src/components/park/parkDates.js
@@ -13,7 +13,7 @@ export const ReservationButtons = ({ campingTypeCode, parkOperation }) => {
     "backcountry-camping": {
       buttons: [
         {
-          label: "Get permit",
+          label: "Register for permit",
           fieldName: "backcountryPermitUrl"
         },
         {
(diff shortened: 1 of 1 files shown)


src/gatsby/src/components/park/parkDates.js
```diff
@@ -46,11 +46,7 @@ export const ReservationButtons = ({ campingTypeCode, parkOperation }) => {
       buttons: [
         {
           label: "Book groupsite",
-          fieldName: "frontcountryGroupReservationUrl"
-        },
-        {
-          label: "Book groupsite",
-          fieldName: "backcountryGroupReservationUrl"
+          fieldName: "frontcountryGroupReservationUrl" || "backcountryGroupReservationUrl"
```
molund, line 49: "frontcountryGroupReservationUrl" is a string so the first half will always be true

ayumi-oxd: Doh yes it is just an object value ... 😅

molund (approved):

PR #1470, opened 2024-09-12
CMS-313: Add visitor guidelines icons

ayumi-oxd:
### Jira Ticket:
CMS-313

### Description:
- Add visitor guidelines icons
- Add SVG files, delete dup SVG files


molund:
Is there a good way to keep track of which font awesome pro icons these are all based on? And which ones are custom or altered?
e.g. a README file in the fontAwesomeIcons folder or an html comment in the individual svg source, or both?

molund:
@ayumi-oxd can you pelase add a readme with a mapping of font awesome icon names to svg file names and some brief instructions describing where we get the SVG icons from?

molund (approved):

PR #1483, opened 2024-10-04
CMS-421: Add filter that removes draft state activities and facilities to park search api

ayumi-oxd:
### Jira Ticket:
CMS-421

### Description:
- Add filter that removes draft state activities and facilities to park search api


duncan-oxd (approved):
Sorry for the delay on this one! I looked at it yesterday but I didn't hit "Approve" 🙈

PR #1485, opened 2024-10-07
CMS-376: Update h5 style

ayumi-oxd:
### Jira Ticket:
CMS-376

### Description:
- Update h5 style

### DIFF
--- a/src/gatsby/src/styles/global.scss
+++ b/src/gatsby/src/styles/global.scss
@@ -83,6 +83,11 @@ h1, h2 {
   color: $colorBlue;
 }
 
+h5 {
+  font-size: 1rem;
+  line-height: normal;
+}
+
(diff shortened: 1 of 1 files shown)


src/gatsby/src/styles/global.scss
```diff
@@ -83,6 +83,11 @@ h1, h2 {
   color: $colorBlue;
 }
 
+h5 {
+  font-size: 1rem;
+  line-height: normal;
```
duncan-oxd, line 88: Jane mentioned the styles for h5 should be the same as normal bold text. On the page she linked to, I think the line-height for the body text is 1.6 but "normal" would make it smaller than that. Probably hard to tell because h5s have margins and stuff, but would it make sense to change this to `line-height: 1.6;`?

duncan-oxd (approved):
Looks good to me!

PR #1487, opened 2024-10-08
CMS-398: Add subArea id to subArea accordion in the camping section

ayumi-oxd:
### Jira Ticket:
CMS-398

### Description:
- Add subArea id to subArea accordion in the camping section

### DIFF
--- a/src/gatsby/src/components/park/parkDates.js
+++ b/src/gatsby/src/components/park/parkDates.js
@@ -202,7 +202,6 @@ export default function ParkDates({ data, parkOperation }) {
     checkHash()
   }, [checkHash])
 
-
   useEffect(() => {
     if (subAreas.length === 1) {
       setOpenAccordions({ 0: true })
--- a/src/gatsby/src/components/park/parkDates.js
+++ b/src/gatsby/src/components/park/parkDates.js
@@ -1,4 +1,5 @@
-import React, { useState, useEffect, useMemo } from "react"
+import React, { useState, useEffect, useMemo, useCallback } from "react"
+import _ from "lodash"
 import Accordion from "react-bootstrap/Accordion"
 import Row from "react-bootstrap/Row"
 import Col from "react-bootstrap/Col"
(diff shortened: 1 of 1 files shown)


src/gatsby/src/components/park/parkDates.js
```diff
@@ -164,6 +171,37 @@ export default function ParkDates({ data, parkOperation }) {
...
 
+  const checkHash = useCallback(() => {
+    // Check hash in url
+    // if we find a matching parkSubArea, open that subArea accordion
+    let h = ""
+    let idx = 0
+    if (typeof window !== "undefined") {
+      h = window.location.hash
+      if (h !== undefined && h !== hash) {
+        subAreas.forEach(subArea => {
```
duncan-oxd, line 182: If it's only searching for one subArea, maybe we should use a find/findIndex instead of a forEach, so it doesn't keep looping unnecessarily after it finds the subArea? (Or is it possible to find multiple matching subAreas?)

duncan-oxd (changes requested):
Just a couple of questions 👍

src/gatsby/src/components/park/parkDates.js
```diff
@@ -164,6 +171,37 @@ export default function ParkDates({ data, parkOperation }) {
...
+            }
+            return true
+          }
+          return false
+        })
+        if (matchingSubArea) {
+          setHash(window.location.hash)
+        }
+      }
+    }
```
duncan-oxd, line 195: I had some ideas for potentially simplifying this even more. Let me know what you think:

```suggestion
    const windowHash = window?.location?.hash
    if (!windowHash || windowHash === hash) return

    const matchingSubAreaIndex = subAreas.findIndex(
      (subArea) => windowHash === "#" + _.kebabCase(subArea.parkSubArea)
    )

    // If we don't find a matching parkSubArea, return
    if (matchingSubAreaIndex === -1) return

    if (!openAccordions[matchingSubAreaIndex]) {
      setOpenAccordions((prev) => ({
        ...prev,
        [matchingSubAreaIndex]: true,
      }))
    }

    setHash(windowHash)
```

I changed `find` to `findIndex`, since the index is what we want, and I moved the call to `setOpenAccordions` after the search, to avoid side effects in the `findIndex` loop.

The rest of it is just "early return" changes to reduce the number of indents. I know not everyone's a fan of those but I find it easier to keep track of what's going on with less nesting.

Hopefully the logic is still the same. Please take a look and let me know if any of these changes might improve readability!

duncan-oxd (commented):
A minor refactoring suggestion to consider, but it looks good either way. Let me know if you have any questions!

src/gatsby/src/components/park/parkDates.js
ayumi-oxd: Yes, it works as expected! Thanks for your advice 🙏

PR #1495, opened 2024-10-17
CMS-129: Change park card layout

ayumi-oxd:
### Jira Ticket:
CMS-129

### Description:
- Change park card layout
- Fix an error `setIsParkOpen is not a function.`

### DIFF
--- a/src/gatsby/src/styles/search.scss
+++ b/src/gatsby/src/styles/search.scss
@@ -310,6 +310,10 @@ div[class$="-singleValue"] {
 }
 
 .park-heading-text {
+  display: -webkit-box;
+  -webkit-line-clamp: 2;
+  -webkit-box-orient: vertical;  
+  overflow: hidden;
   color: $colorBlue;
   font-size: 1.25rem;
   line-height: 2rem;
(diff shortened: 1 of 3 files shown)


src/gatsby/src/styles/search.scss
```diff
@@ -310,6 +310,10 @@ div[class$="-singleValue"] {
 }
 
 .park-heading-text {
+  display: -webkit-box;
+  -webkit-line-clamp: 2;
+  -webkit-box-orient: vertical;
```
duncan-oxd, line 315: :O I don't know anything about these CSS rules but I'm reading that they're supported in Edge/Firefox/Chrome 👍 
If this project is set up to auto-prefix styles, maybe there's a way to write it without the `-webkit` prefixes?

(Looks like maybe there isn't, since the rules are still a working draft...?)

duncan-oxd (approved):
Looks good; interesting cutting-edge CSS rules 😮 
We may have to keep an eye on it if the spec changes for those CSS rules, but for now it looks like it works 👍

src/gatsby/src/styles/search.scss
ayumi-oxd: Yup, we don't have it ... but it'd be a good idea to have a rule at some point.

PR #1498, opened 2024-10-22
CMS-377: Display an event type in the select field based on the url

ayumi-oxd:
### Jira Ticket:
CMS-377

### Description:
- Display an event type in the select field based on the url

### DIFF
--- a/src/gatsby/src/components/advisories/advisoryFilter.js
+++ b/src/gatsby/src/components/advisories/advisoryFilter.js
@@ -1,4 +1,5 @@
-import React, { useState } from "react"
+import { navigate } from "gatsby"
+import React, { useState, useEffect } from "react"
 import Form from "react-bootstrap/Form"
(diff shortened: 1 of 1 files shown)


src/gatsby/src/components/advisories/advisoryFilter.js
```diff
@@ -91,21 +105,27 @@ const AdvisoryFilter = ({
             id="event-search-typeahead"
             minLength={0}
             labelKey="label"
+            filterBy={() => true}
+            options={eventTypes}
+            selected={selectedEventType}
+            onChange={(selected) => handleOnChange(selected)}
```
duncan-oxd, line 111: I might call this method `handleTypeaheadChange` in this case, since the "on" is kind of redundant

duncan-oxd (approved):
I had one suggestion about a variable name, but the rest looks good 👍

PR #1501, opened 2024-10-24
CMS-129: Update park card layout

ayumi-oxd:
### Jira Ticket:
CMS-129

### Description:
- Update park card layout

### DIFF
--- a/src/gatsby/src/styles/search.scss
+++ b/src/gatsby/src/styles/search.scss
@@ -310,26 +310,27 @@ div[class$="-singleValue"] {
 }
 
 .park-heading-text {
-  display: -webkit-box;
-  -webkit-line-clamp: 2;
-  -webkit-box-orient: vertical;  
-  overflow: hidden;
+  display: inline-flex;
+  position: relative;
   color: $colorBlue;
   font-size: 1.25rem;
   line-height: 2rem;
   margin-bottom: 4px;
   a.desktop-park-link,
   a.mobile-park-link {
-    position: relative;
+    display: -webkit-box;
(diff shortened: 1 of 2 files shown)


ayumi-oxd:
@duncan-oxd The second park card, the arrow icon after the park name is truncated. Using `absolute` doesn't work... 
Would you happen to have any ideas for this fix?

Screenshot 2024-10-23 at 5 46 24 PM

duncan-oxd:
Maybe if you made the text container a flex box, so you could control the size of the text element and always have the park-heading-icon right after it?

src/gatsby/src/styles/search.scss
```diff
@@ -310,19 +310,27 @@ div[class$="-singleValue"] {
 }
 
 .park-heading-text {
-  display: -webkit-box;
-  -webkit-line-clamp: 2;
-  -webkit-box-orient: vertical;  
-  overflow: hidden;
+  display: inline-flex;
+  position: relative;
```
duncan-oxd, line 314: If we want to center the text and the icon vertically, we can add `align-items: center;` to the container - then we can remove the `position: relative;` and all of the styles on the `.park-heading-icon` except the (font size).
```suggestion
  display: inline-flex;
  align-items: center;
  gap: 0.5em; // some spacing to keep the distance between the icon and the text - not sure how much
```

I think the result will be the same, but more robust because flex box will handle all the positioning.

ayumi-oxd: Thanks for your suggestion, works like a charm 👏

duncan-oxd (approved):

PR #1503, opened 2024-10-24
CMS-363: Update snowplow tracking

ayumi-oxd:
### Jira Ticket:
CMS-363

### Description:
- Update snowplow tracking


duncan-oxd (approved):

molund:
The idea behind this ticket was actually to call the snowplow event every time a page renders or re-renders, not just when the router is updated.  For example if you use the pagination control at the bottom of the active advisories page the url doesn't change but new content is added to the page which is not hooked up to any snowplow events

duncan-oxd:
Ah... Hopefully it's as simple as adding tracking call to a few more event hooks

PR #1504, opened 2024-10-25
CMS-363: Update snowplow tracking

ayumi-oxd:
### Jira Ticket:
CMS-363

### Description:
- Added `useEffect` hook to call the snowplow event every time a page renders or re-renders, not just when the router is updated, based on Mike's comment: https://github.com/bcgov/bcparks.ca/pull/1503
- Used `wrapPageElement` to wrap all pages with the custom hook: https://www.gatsbyjs.com/docs/reference/config-files/gatsby-browser/#wrapPageElement
- Need to test it on the DEV/TEST/PROD since we have `GATSBY_ENABLE_SNOWPLOW` in the `.env`

### DIFF
--- a/src/gatsby/gatsby-browser.js
+++ b/src/gatsby/gatsby-browser.js
@@ -8,6 +8,10 @@ import "./src/styles/style.scss"
 import React from "react"
 import useSnowplowTracking from "./src/utils/useSnowplowTracking"
 
+export const onRouteUpdate = ({ location, prevLocation }) => {
+  sessionStorage.setItem("prevPath", prevLocation ? prevLocation.pathname : null);
+}
+
 const SnowplowWrapper = ({ element }) => {
   useSnowplowTracking()
   return element
--- a/src/gatsby/gatsby-browser.js
+++ b/src/gatsby/gatsby-browser.js
@@ -5,14 +5,17 @@ import "bootstrap/dist/js/bootstrap.bundle"
 import "@fortawesome/fontawesome-free/css/all.min.css"
 import "./src/styles/style.scss"
 
-export const onRouteUpdate = ({ location, prevLocation }) => {
-  if (typeof window.snowplow === 'function') {
-    window.snowplow("trackPageView");
-    // refresh snowplow link click tracking to pick up new links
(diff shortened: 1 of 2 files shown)


src/gatsby/gatsby-browser.js
```diff
@@ -5,14 +5,17 @@ import "bootstrap/dist/js/bootstrap.bundle"
 import "@fortawesome/fontawesome-free/css/all.min.css"
 import "./src/styles/style.scss"
 
-export const onRouteUpdate = ({ location, prevLocation }) => {
-  if (typeof window.snowplow === 'function') {
-    window.snowplow("trackPageView");
-    // refresh snowplow link click tracking to pick up new links
-    window.snowplow("refreshLinkClickTracking");
-  }
-  sessionStorage.setItem("prevPath", prevLocation ? prevLocation.pathname : null);
```
duncan-oxd, line 14: is `prevPath` in sessionStorage not needed anymore?

ayumi-oxd: Good catch! Accidentally deleted it 😅

src/gatsby/gatsby-browser.js
```diff
@@ -5,14 +5,21 @@ import "bootstrap/dist/js/bootstrap.bundle"
...
+}
+
+const SnowplowWrapper = ({ element }) => {
+  useSnowplowTracking()
+  return element
+}
+
+export const wrapPageElement = ({ element }) => {
+  return <SnowplowWrapper element={element} />
+}
```
duncan-oxd, line 22: Do we need to add this to the template, or is it an automatic thing in Gatsby?

ayumi-oxd: If I understand correctly, it's an automatic thing. It allows you to wrap the root element of every page: https://www.gatsbyjs.com/docs/reference/config-files/gatsby-browser/#wrapPageElement
https://www.gatsbyjs.com/docs/reference/config-files/gatsby-browser/#usage
If you add `console.log()` in the `useSnowplowTracking`, you can see it when a page is loaded.

duncan-oxd (approved):

PR #1510, opened 2024-11-01
CMS-363: Add snowplow track to onRouteUpdate

ayumi-oxd:
### Jira Ticket:
CMS-363

### Description:
- Add snowplow track to onRouteUpdate
- Revert the initial idea since it was needed for single-page applications like Gatsby, where navigation between pages does not trigger a full-page reload

### DIFF
--- a/src/gatsby/gatsby-browser.js
+++ b/src/gatsby/gatsby-browser.js
@@ -9,6 +9,10 @@ import React from "react"
 import useSnowplowTracking from "./src/utils/useSnowplowTracking"
 
 export const onRouteUpdate = ({ location, prevLocation }) => {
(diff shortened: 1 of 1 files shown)


duncan-oxd:
If you open a fresh tab and navigate to a page, will that fire `onRouteUpdate`? If not, do we need to also add any tracking to a hook like `onInitialClientRender`?

ayumi-oxd:
> If you open a fresh tab and navigate to a page, will that fire `onRouteUpdate`? If not, do we need to also add any tracking to a hook like `onInitialClientRender`?

In that case `useSnowplowTracking` will do the job since it will run on mount and unmount.

duncan-oxd (approved):

PR #1513, opened 2024-11-06
CMS-461: Fix incorrect advisory count displayed on Park Operating dates page

ayumi-oxd:
### Jira Ticket:
CMS-461

### Description:
- Fix incorrect advisory count displayed on Park Operating dates page
- The issue happened because we didn't get the advisories data

### DIFF
--- a/src/gatsby/src/pages/plan-your-trip/park-operating-dates.js
+++ b/src/gatsby/src/pages/plan-your-trip/park-operating-dates.js
@@ -60,10 +60,6 @@ const ParkLink = ({ park, apiBaseUrl }) => {
   }
 
   // get advisories
-  const getAccessStatus = (status) => {
-    setParkAccessStatus(status)
-  }
-
   useEffect(() => {
     setIsLoadingAdvisories(true)
     loadAdvisories(apiBaseUrl, park.orcs)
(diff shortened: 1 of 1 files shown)


src/gatsby/src/pages/plan-your-trip/park-operating-dates.js
```diff
@@ -54,6 +59,42 @@ const ParkLink = ({ park, advisories }) => {
     }
   }
 
+  // get advisories
+  const getAccessStatus = (status) => {
+    setParkAccessStatus(status)
+  }
```
duncan-oxd, line 65: Should this name be "set" instead of "get" ? And if so, couldn't you just call setParkAccessStatus without this other function?

src/gatsby/src/pages/plan-your-trip/park-operating-dates.js
```diff
@@ -54,6 +59,42 @@ const ParkLink = ({ park, advisories }) => {
...
+    setIsLoadingAdvisories(true)
+    loadAdvisories(apiBaseUrl, park.orcs)
+      .then(response => {
+        if (response.status === 200) {
+          setAdvisories(response.data.data)
+          setAdvisoryLoadError(false)
+        } else {
+          setAdvisories([])
+          setAdvisoryLoadError(true)
+        }
```
duncan-oxd, line 77: If the API sends back an error code like 404 or 500, the axios call in `loadAdvisories` will throw an exception, so those errors would go to a `.catch` block instead of the `.then`, I think!

duncan-oxd (approved):

PR #1517, opened 2024-11-08
CMS-445: Update logic to display park gate open/close time

ayumi-oxd:
### Jira Ticket:
CMS-445

### Description:
- Update logic to display park gate open/close time

### DIFF
--- a/src/gatsby/src/components/park/parkHeader.js
+++ b/src/gatsby/src/components/park/parkHeader.js
@@ -48,13 +48,36 @@ const mapImages = {
 }
 // Helper function to format gate open/close time e.g. "08:00:00" to "8 am"
 const formattedTime = (time) => {
-  // prepend a dummy date to the time string to parse it
+  // Prepend a dummy date to the time string to parse it
   const dateTime = parseISO(`1970-01-01T${time}`)
   const minutes = format(dateTime, 'mm');
   if (minutes === '00') {
     return format(dateTime, 'h aa').toLowerCase()
   } else {
     return format(dateTime, 'h:mm aa').toLowerCase()
+  }
+}
+// Helper function to render the gate open/close times
+const renderGateTimes = (parkOperation) => {
+  if (!parkOperation) return null
+  const { gateOpenTime, gateCloseTime, gateOpensAtDawn, gateClosesAtDusk } = parkOperation
+
+  if (gateOpenTime === "00:00:00" && gateCloseTime === "23:59:00") {
+    return <>24 hours a day.</>
+    // Either gateOpenTime or gateOpensAtDawn is available, then either gateCloseTime or gateClosesAtDusk must also be available
+  } else if ((gateOpenTime || gateOpensAtDawn) && (gateCloseTime || gateClosesAtDusk)) {
+    return (
+      <>
+        , from <span className="no-wrap">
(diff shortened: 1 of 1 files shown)


src/gatsby/src/components/park/parkHeader.js
```diff
@@ -130,16 +130,16 @@ export default function ParkHeader({
...
-                  {(parkOperation?.gateOpenTime && parkOperation?.gateCloseTime) ? (
-                    parkOperation.gateOpenTime === "00:00:00" && parkOperation.gateCloseTime === "23:59:00" ? (
+                  {(parkOperation?.gateOpenTime || parkOperation?.gateCloseTime || parkOperation?.gateOpensAtDawn || parkOperation?.gateClosesAtDusk) ? (
+                    parkOperation?.gateOpenTime === "00:00:00" && parkOperation?.gateCloseTime === "23:59:00" ? (
                       <>, 24 hours a day.</>
                     ) : (
                       <>
                         , from <span className="no-wrap">
-                          {parkOperation?.gateOpensAtDawn ? "dawn" : formattedTime(parkOperation.gateOpenTime)}
+                          {parkOperation?.gateOpensAtDawn ? "dawn" : formattedTime(parkOperation?.gateOpenTime)}
```
duncan-oxd, line 139: If `parkOperation.gateOpenTime` will ever be null/undefined (I see the added `?`) then it won't render anything and it might say "..is open  to  , daily."

Is there any more conditional logic we should sort out in the code before we print it in the template?

duncan-oxd: Looking at the big block of conditions on lines 133-134, maybe we should move all of that up into the code and set a variable or two for the gate times to put into the template?

ayumi-oxd: My bad 😅 The logic is that if either `gateOpenTime` or `gateOpensAtDawn` is available, then either `gateCloseTime` or `gateClosesAtDusk` must also be available. And if one of them is missing, display a period ".".

src/gatsby/src/components/park/parkHeader.js
```diff
@@ -130,20 +153,7 @@ export default function ParkHeader({
...
-                        , from <span className="no-wrap">
-                          {parkOperation?.gateOpensAtDawn ? "dawn" : formattedTime(parkOperation.gateOpenTime)}
-                        </span>{" "}
-                        to <span className="no-wrap">
-                          {parkOperation?.gateClosesAtDusk ? "dusk" : formattedTime(parkOperation.gateCloseTime)}
-                        </span>, daily.
-                      </>
-                    )
-                  ) : "."}
+                  {renderGateTimes(parkOperation)}
```
duncan-oxd, line 156: Nice, I think that function is a big improvement for readability. One minor thing I notice: there's whitespace here between `{parkDates}` and the function call in the template, so it will render with a space. That's good if it returns "24 hours a day" etc, but if it just returns a `.` you won't want the space before the period. 

We might need to do a conditional `{" "}` in the template to handle that.

ayumi-oxd: It doesn't seem to return the space before the period on my end, but let me know if you see it 🙂
Screenshot 2024-11-12 at 1 26 13 PM

duncan-oxd (approved):

PR #1520, opened 2024-11-13
CMS-472: Update park contact lifecycle in Strapi

ayumi-oxd:
### Jira Ticket:
CMS-472

### Description:
- Update park contact lifecycle in Strapi

### DIFF
--- a/src/cms/src/api/park-contact/content-types/park-contact/lifecycles.js
+++ b/src/cms/src/api/park-contact/content-types/park-contact/lifecycles.js
@@ -6,12 +6,24 @@ const updateName = async (data, where) => {
         const parkContact = await strapi.entityService.findOne(
             "api::park-contact.park-contact", id, { populate: '*' }
         )
-        const protectedArea = parkContact.protectedArea
+        let protectedArea = parkContact.protectedArea
         const parkOperatorContact = parkContact.parkOperatorContact
+
+        // Check if new protectedArea is being added
+        if (data?.protectedArea?.connect?.length > 0) {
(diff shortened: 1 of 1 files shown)


src/cms/src/api/park-contact/content-types/park-contact/lifecycles.js
```diff
@@ -6,12 +6,24 @@ const updateName = async (data, where) => {
...
+        } else if (data?.protectedArea?.disconnect?.length > 0) {
+          protectedArea = { orcs: 0 }
+        }
+
         data.name = ""
         if (protectedArea) {
             data.name = protectedArea.orcs
+        } else {
+            data.name = 0
         }
```
duncan-oxd, line 27: should this be checking the orcs? If disconnect.length > 0 and you set the orcs to 0, `protectedArea` will still evaluate to true in this check. Just making sure that's intended.

ayumi-oxd: Yes, that's intended. If `parkContact` is created for the first time without `protectedArea`, it'll fall in this `else {}`. However once users `protectedArea` add to `parkContact` and remove `protectedArea` from it, and click the Save button, it'll still return the previous `protectedArea` (I guess it's because of how Strapi lifecycle works) 😅 so I needed to change its `orcs` when it's removed (disconnected).

duncan-oxd: Thanks for clarifying!

duncan-oxd (approved):

PR #1529, opened 2024-11-20
CMS-373: Display updated date or posted date if 'No date' is selected

ayumi-oxd:
### Jira Ticket:
CMS-373
CMS-374
CMS-403

### Description:
- Display updated date or posted date if 'No date' is selected
- Fix the pills layout issue on the active advisories page
- Update display date logic on park page - reuse `AdvisoryDate` component


src/gatsby/src/components/advisories/advisoryCard.js
```diff
@@ -248,6 +248,22 @@ const AdvisoryCard = ({ advisory, parkInfoHash }) => {
...
+                </div>
+              ) : (
+                formatDate(advisory.advisoryDate) && (
+                  <div className="date">
+                    <small>Posted</small>
+                    <small><b>{formatDate(advisory.advisoryDate)}</b></small>
+                  </div>
+                )
+              )
+            )}
```
duncan-oxd, line 266: I always find blocks like these to be hard to read 😓 Chained `&&`s and a multi-line ternary block with more `&&`s. It works the way it is, but it might be a big win for readability to move some of that logic out of the template. 

Maybe have a variable like "`noDateSelected`" and/or a function to handle this logic and return the appropriate block of JSX? 

You don't have to change anything here, but that's something to think about next time, if you find yourself adding a lot of logic in the template 👍

duncan-oxd (approved):
Looks good! I put in a note about readability but it's not required to change here 👍

src/gatsby/src/components/advisories/advisoryCard.js
ayumi-oxd: My bad, I agree with you 😅 Yeap it has so many conditions ... Updated my PR, hope it got better 🙏

src/gatsby/src/styles/advisories/advisoryDate.scss
ayumi-oxd, line 1: This indent looks different from my end for some reason ...

duncan-oxd: It looks like this file is using tab indents while all of the other files are using spaces. Let's convert this to spaces and re-save it

PR #1533, opened 2024-11-23
CMS-400: Fix the issue when navigating page section directly from a url

ayumi-oxd:
### Jira Ticket:
CMS-400

### Description:
- Fix the issue when navigating page section directly from a url
- I'm not 100% sure if it'll fix it since I couldn't recreate it on my local 🤔 Just applied the change based on their doc: https://www.npmjs.com/package/react-use-scrollspy

### DIFF
--- a/src/gatsby/src/templates/park.js
+++ b/src/gatsby/src/templates/park.js
@@ -125,7 +125,9 @@ export default function ParkTemplate({ data }) {
       const id = window.location.hash.replace("#", "")
       const element = document.getElementById(id) || document.querySelector(window.location.hash)
       if (element) {
-        element.scrollIntoView({ behavior: 'smooth', block: 'start' })
+        setTimeout(() => {
+          element.scrollIntoView({ behavior: 'smooth', block: 'start' })
+        }, 100)
       }
     }
   }, [isLoadingProtectedArea, isLoadingAdvisories])
--- a/src/gatsby/src/templates/park.js
+++ b/src/gatsby/src/templates/park.js
@@ -134,15 +134,15 @@ export default function ParkTemplate({ data }) {
     setParkAccessStatus(statusObj);
   };
 
-  const parkOverviewRef = useRef("")
-  const knowBeforeRef = useRef("")
-  const mapLocationRef = useRef("")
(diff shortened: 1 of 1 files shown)


duncan-oxd:
From the video Manuji posted in the ticket, it looks like the browser doesn't quite scroll far enough to trigger the "Contact" section being highlighted. (I would expect it to scroll "Contact" to the top of the screen, but it stops half way up the page. It might be related to React Router, or: is there any special code that handles that scrolling to specific page sections?

ayumi-oxd:
> From the video Manuji posted in the ticket, it looks like the browser doesn't quite scroll far enough to trigger the "Contact" section being highlighted. (I would expect it to scroll "Contact" to the top of the screen, but it stops half way up the page. It might be related to React Router, or: is there any special code that handles that scrolling to specific page sections?

We have this scroll in `useEffect`,
```
  useEffect(() => {
    if (window.location.hash && !isLoadingProtectedArea && !isLoadingAdvisories) {
      const id = window.location.hash.replace("#", "")
      const element = document.getElementById(id) || document.querySelector(window.location.hash)
      if (element) {
        element.scrollIntoView({ behavior: 'smooth', block: 'start' })
      }
    }
  }, [isLoadingProtectedArea, isLoadingAdvisories])
```
And this `useScrollSpy` to highlight the side menu based on the scroll position.
```
  const activeSection = useScrollSpy({
    sectionElementRefs: sectionRefs,
    defaultValue: 0,
    offsetPx: -100,
  })
```
So I might need to take a look at `useEffect` part more. The tricky part is that I couldn't recreate it locally 😅

duncan-oxd:
> So I might need to take a look at `useEffect` part more. The tricky part is that I couldn't recreate it locally 😅

Chrome seems unreliable with scrollIntoView when a page is loading or an element is just mounting. I noticed this a lot on previous projects. I hate to say it, but the only solution at the time was to pause for a few milliseconds, or run scrollIntoView twice. Both ways felt like bad hacks :( Let me know if you find a cleaner solution!

duncan-oxd (approved):
Hopefully the setTimeout is enough to make it work reliably enough for most users!

PR #1534, opened 2024-11-26
CMS-473: Add no results messaging

ayumi-oxd:
### Jira Ticket:
CMS-473

### Description:
- Update no results messaging
- Add no results messaging to the dates page, A-Z list page, and the approved page

### DIFF
--- a/src/gatsby/src/components/search/noSearchResults.js
+++ b/src/gatsby/src/components/search/noSearchResults.js
@@ -2,93 +2,133 @@ import React from "react"
 import { Link } from "gatsby"
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome"
 import { faSearch, faArrowUpRightFromSquare } from "@fortawesome/free-solid-svg-icons"
+import PropTypes from "prop-types"
 
 import "../../styles/search.scss"
 
+const MapLink = () => {
+  return (
+    <a
+      className="map-link"
+      href="https://governmentofbc.maps.arcgis.com/apps/webappviewer/index.html?id=077ef73a1eae4ca88f2bafbb831215af&query=British_Columbia_Parks_Ecological_Reserves_and_Protected_Areas_8747,ORCS_PRIMARY,0000"
+    >
+      Map of parks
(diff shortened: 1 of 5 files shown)


src/gatsby/src/components/search/noSearchResults.js
```diff
@@ -2,93 +2,133 @@ import React from "react"
...
-              aria-label="Remove city name"
-              onClick={handleClickClearCity}
-              onKeyDown={handleKeyDownClearCity}
+      {page === "find-a-park" && (
+        <>
+          <p><b>You could:</b></p>
+          <ul>
+            <li>Try a different search term</li>
+            <li>
+              <button
```
duncan-oxd, line 68: You can "press" a button with a mouse click, or by focusing it and pressing space (or enter) but I _think_ all 3 of these event swill fire a `click` event, so you don't need to specifically handle onKeyDown. (Double check if that's true in React, though...)

src/gatsby/src/components/search/noSearchResults.js
```diff
@@ -2,93 +2,133 @@ import React from "react"
...
+      )}
+      {(page === "park-operating-dates" || page === "a-z-list") && (
+        <>
+          <p className="mb-4">
+            <b>Select 'All' to go back to the full list.</b>
+          </p>
+          <p><b>Or search using the:</b></p>
+          <ul>
+            <li>
+              <Link to="/find-a-park">Find a park</Link> page
```
duncan-oxd, line 98: I remember we updated the links in #1531 to include "page" in the anchor text; do we want to do that here too? I see "...list" is already included in the A-Z list link.

src/gatsby/src/components/search/noSearchResults.js
ayumi-oxd: Good catch! Yes it works in React as well.

src/gatsby/src/components/search/noSearchResults.js
ayumi-oxd: Thanks! Brett prefers to include "page" as a link.

duncan-oxd (approved):

PR #1537, opened 2024-11-26
CMS-511: Remove the search result count line

ayumi-oxd:
### Jira Ticket:
CMS-511

### Description:
- Remove the search result count line since it's a dup
- No need `my-2` if there is no text

### DIFF
--- a/src/gatsby/src/pages/active-advisories.js
+++ b/src/gatsby/src/pages/active-advisories.js
@@ -384,8 +384,8 @@ const PublicActiveAdvisoriesPage = ({ data }) => {
         </div>
 
         <div className={isDataOld ? "hidden" : undefined}>
-          {isSearchError &&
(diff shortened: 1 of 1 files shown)


src/gatsby/src/pages/active-advisories.js
duncan-oxd, line 1: If it's going to be hidden anyway, would it make sense to just leave the whole element out if `isAnySearch` is false?

duncan-oxd (changes requested):
Just a question to potentially simplify this even more 👍

src/gatsby/src/pages/active-advisories.js
ayumi-oxd: Haha exactly, good catch!

duncan-oxd (approved):

PR #1538, opened 2024-11-28
CMS-206: Update bootstrap to v5

ayumi-oxd:
### Jira Ticket:
CMS-206

### Description:
- Update bootstrap to v5
  - Change `.no-gutters` to `.g-0`
  - Change `.badge-pill` to `.rounded-pill`
  - Change `.text-left` to `.text-start`
  - Change `.ml-*` and `.mr-*` to `.ms-*` and `.me-*`
  - Change `.pl-*` and `.pr-*` to `.ps-*` and `.pe-*`
  - Change `.sr-only` and `.sr-only-focusable` to `.visually-hidden` and `.visually-hidden-focusable`
  - Update `media-breakpoint-down` since it uses the breakpoint itself instead of the next breakpoint 
  - Remove `.form-group` styling since it's dropped
  - Remove `font-weight-bold` class since it's dropped
- Update gov bootstrap to v5
- Update react-bootstrap to v2, which used bootstrap v5
  - Remove `Accordion.Toggle` since it's dropped


src/gatsby/src/archived/about.js
duncan-oxd, line 1: Were these two archived files meant to be deleted in this branch?

ayumi-oxd: I added these files in the archived folder a half year ago: https://github.com/bcgov/bcparks.ca/tree/main/src/gatsby/src/archived but now I don't think it'll be useful in the future 😅 Also keeping them can be additional work when it comes to library updates ...

duncan-oxd: Fair enough! Just making sure it was intentional to delete them 👍

duncan-oxd (approved):
The code changes look sensible, but QA knows what the site should look like better than me 😁 
I commented about a few things to consider, but it's ready to merge 👍

PR #1540, opened 2024-11-29
CMS-461: Hide the check advisory link if the park doesn't have an advisory

ayumi-oxd:
### Jira Ticket:
CMS-461

### Description:
- Hide the check advisory link if the park doesn't have an advisory


duncan-oxd (approved):
You could probably re-order it to only need one check on `advisories.length` but it looks like it'll work 👍

PR #1543, opened 2024-12-05
CMS-473: Update no results styling

ayumi-oxd:
### Jira Ticket:
CMS-473

### Description:
- Update no results styling
- Fix that bullet points were missing in the No result message

### DIFF
--- a/src/gatsby/src/styles/search.scss
+++ b/src/gatsby/src/styles/search.scss
@@ -582,6 +582,12 @@ div[class$="-singleValue"] {
       outline: $colorBlueMed auto 1px;
     }
   }
+  ul {
+    padding-left: 32px !important;
+    li {
+      list-style-type: disc !important;
+    }
+  }
 }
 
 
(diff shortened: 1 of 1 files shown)


src/gatsby/src/styles/search.scss
```diff
@@ -582,6 +582,12 @@ div[class$="-singleValue"] {
       outline: $colorBlueMed auto 1px;
     }
   }
+  ul {
+    padding-left: 32px !important;
```
duncan-oxd, line 586: The style from commons.css on the homepage has this as `2rem` and it might be good to use rem here too, so it will scale with the font size. (If that ever happens)

By the way, when I tested this locally, there were parks with every letter including X/Y/Z - did you manually filter something out to test this?

ayumi-oxd: I just added `<NoSearchResults />` outside of the condition and checked how it looked like 😂 Changing park names/losing parks is not common, so this message probably won't be seen on the A-Z page but it's always good to have a fallback!

duncan-oxd (approved):

PR #1545, opened 2024-12-09
CMS-544: Fix the load more button on Active advisory page takes the user back to top of the page

ayumi-oxd:
### Jira Ticket:
CMS-544

### Description:
- Fix the issue that the load more button on Active advisory page takes the user back to top of the page
- The previous way caused the page re-rendering - removed `pageIndex` from `getAdvisories` useCallback dependency

### DIFF
--- a/src/gatsby/src/pages/active-advisories.js
+++ b/src/gatsby/src/pages/active-advisories.js
@@ -224,11 +224,10 @@ const PublicActiveAdvisoriesPage = ({ data }) => {
   const getAdvisories = useCallback(
     q => {
       // q = api query
-
-      let newApiCall = apiUrl + `/public-advisories` + q
-
-      newApiCall += "&limit=" + pageLen // use -1 for unlimited
-      newApiCall += "&start=" + pageLen * (pageIndex - 1)
+      const params = new URLSearchParams(q)
+      params.append("limit", pageLen)
+      params.append("start", pageLen * (pageIndex - 1))
+      const newApiCall = `${apiUrl}/public-advisories?${params.toString()}`
 
       if (apiCall !== newApiCall) {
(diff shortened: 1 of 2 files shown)


src/gatsby/src/pages/active-advisories.js
```diff
@@ -282,12 +282,30 @@ const PublicActiveAdvisoriesPage = ({ data }) => {
...
+    setPageIndex(newIndex)
+    const pageStart = (newIndex - 1) * pageLen
+
+    const aType = getAdvisoryTypeFromUrl()
+    let q = getApiQuery(aType)
+
+    let newApiCall = apiUrl + `/public-advisories` + q
+    newApiCall += "&limit=" + pageLen
+    newApiCall += "&start=" + pageStart
+    axios.get(newApiCall).then(resultResponse => {
```
duncan-oxd, line 300: This came up in another PR recently, so I'll mention it here too: You might find the URLSearchParams API a bit easier to work with in the future https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams compared to manually building strings and managing the "&" characters 👍

duncan-oxd (approved):

src/gatsby/src/pages/active-advisories.js
ayumi-oxd: Thanks! I didn't know this but it looks very useful 😄 I'd appreciate it if you could take a second look 🙏

duncan-oxd: Looks good! There's also the URL API for the main part of the URL, but it's not _as_ useful unless you need to parse out or change a part of an existing url: https://developer.mozilla.org/en-US/docs/Web/API/URL/URL 🎓

PR #1546, opened 2024-12-09
CMS-545: Add retry to fetch advisories on park operating dates page

ayumi-oxd:
### Jira Ticket:
CMS-545

### Description:
- Found 429 errors happen only on TEST env when fetching the advisories data (e.g. https://bcparks.test.api.gov.bc.ca/api/public-advisories/items?filters[protectedAreas][orcs][$eq]=6197&pagination[limit]=100) but not on other envs or my local. Added retry with `setTimeOut` when it has an error to fix this issue. Any advice is more than welcome since I'm not really sure how to fix it 😅

### DIFF
--- a/src/gatsby/src/pages/plan-your-trip/park-operating-dates.js
+++ b/src/gatsby/src/pages/plan-your-trip/park-operating-dates.js
@@ -87,6 +87,7 @@ const ParkLink = ({ park, apiBaseUrl }) => {
   // get advisories
   useEffect(() => {
     fetchAdvisoriesWithRetry()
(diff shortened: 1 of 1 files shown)


duncan-oxd:
I get errors intermittently from Strapi on the dev environment when I'm trying to build the site. It doesn't print out the error code but I bet they're also 429 "Too many requests" 

Let's ask around and see if we can have the request limit raised a bit on these environments so we won't get the error anymore. Or at least not as often!

duncan-oxd (approved):
As discussed on slack, the real fix for this will come from the Strapi server config, but this incremental backoff could still be useful if you want to merge it 👍

PR #1550, opened 2024-12-11
CMS-559: Add time zone to emergency-alert time fields

ayumi-oxd:
### Jira Ticket:
CMS-559

### Description:
- Add time zone to emergency-alert time fields

### DIFF
--- a/src/cms/src/api/emergency-alert/content-types/emergency-alert/lifecycles.js
+++ b/src/cms/src/api/emergency-alert/content-types/emergency-alert/lifecycles.js
@@ -1,18 +1,22 @@
 "use strict";
 
+const formatDateToPacificTime = (dateString) => {
+  const date = new Date(dateString)
+  const localDate = date.toLocaleDateString('en-CA', { timeZone: 'America/Los_Angeles' })
+  return new Date(localDate).toISOString().split('T')[0]
+}
+
 module.exports = {
     beforeCreate(event) {
         const { data } = event.params
-        const date = new Date(data.createdAt)
-        const createdDate = date.toISOString().split('T')[0]
+        const createdDate = formatDateToPacificTime(data.createdAt)
         if (data.isActive === true) {
             data.activeDate = createdDate
         }
     },
     beforeUpdate(event) {
         const { data } = event.params
-        const date = new Date(data.updatedAt)
-        const updatedDate = date.toISOString().split('T')[0]
+        const updatedDate = formatDateToPacificTime(data.updatedAt)
         if (data.isActive === true) {
             data.activeDate = updatedDate
         }
--- a/src/cms/src/api/emergency-alert/content-types/emergency-alert/lifecycles.js
+++ b/src/cms/src/api/emergency-alert/content-types/emergency-alert/lifecycles.js
@@ -1,9 +1,12 @@
 "use strict";
 
+const format = require('date-fns/format')
+
 const formatDateToPacificTime = (dateString) => {
-  const date = new Date(dateString)
-  const localDate = date.toLocaleDateString('en-CA', { timeZone: 'America/Los_Angeles' })
-  return new Date(localDate).toISOString().split('T')[0]
+  // Strapi returns the date in ISO format e.g. 2025-01-01T00:00:00.000Z
+  // Format the date in YYYY-MM-DD e.g. 2025-01-01
+  const formattedDate = format(dateString, 'yyyy-MM-dd', { timeZone: 'America/Los_Angeles' })
+  return formattedDate
 }
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


src/cms/src/api/emergency-alert/content-types/emergency-alert/lifecycles.js
```diff
@@ -1,23 +1,27 @@
 "use strict";
 
+const formatDateToPacificTime = (dateString) => {
+  const date = new Date(dateString)
+  const localDate = date.toLocaleDateString('en-CA', { timeZone: 'America/Los_Angeles' })
+  return new Date(localDate).toISOString().split('T')[0]
```
duncan-oxd, line 6: this `toLocaleDateString` function already returns a YYYY-MM-DD date string; can't we just return that without needing this last line that parses it and formats it again?

ayumi-oxd: This `localDate` returns MM/DD/YYY, like `12/11/2024` so it still needs to be converted 😓

duncan-oxd: In my browser, `en-CA` returns YYYY-MM-DD but `en-US` returns MM/DD/YYYY 🤔 Apparently that can change(??) Maybe because I've got my machine set to use YYYY-MM-DD date formats in the OS? Since this project is already using `date-fns`, you could use that here to get a predictable output (see parkAccessStatus.js where date-fns `format` function is used to output a date with year-month-day format)

duncan-oxd: also, what format is the `dateString` parameter when it gets passed in? Let's add some comments to explain what this function does

ayumi-oxd: Thanks! Using `date-fns` seems to be a good idea. Strapi returns an ISO date string like `2025-01-01T00:00:00.000Z`, but it's interesting I needed to convert it into YYYY-MM-DD format when updating data in the Strapi lifecycle.

src/cms/src/api/emergency-alert/content-types/emergency-alert/lifecycles.js
```diff
@@ -1,23 +1,29 @@
 "use strict";
 
+const format = require('date-fns/format')
+
+const formatDateToPacificTime = (dateString) => {
+  // Strapi returns the date in ISO format e.g. 2025-01-01T00:00:00.000Z
+  // Format the date in YYYY-MM-DD e.g. 2025-01-01
+  return format(dateString, 'yyyy-MM-dd', { timeZone: 'America/Los_Angeles' })
```
duncan-oxd, line 8: I don't think `timeZone` is a valid property for the options object ( https://date-fns.org/v2.28.0/docs/format#arguments ?) but it will be using the local timezone anyway. If we need to specify the timezone, there's a plugin for date-fns we could use (or the newer version 4 apparently has timezone functions built in)

src/cms/src/api/emergency-alert/content-types/emergency-alert/lifecycles.js
```diff
@@ -1,23 +1,32 @@
...
 
+const format = require('date-fns/format')
+const utcToZonedTime = require('date-fns-tz/utcToZonedTime')
+
+const formatDateToPacificTime = (dateString) => {
+  // Strapi returns the date in ISO format e.g. 2025-01-01T00:00:00.000Z
+  // Convert dateString to Pacific Time
+  const pacificTime = utcToZonedTime(dateString, 'America/Los_Angeles')
+  // Format the date in YYYY-MM-DD e.g. 2025-01-01
+  return format(pacificTime, 'yyyy-MM-dd')
```
duncan-oxd, line 11: I haven't used `date-fns-tz` before but it looks like the function we really want is `formatInTimezone` which would let us do this all on one line: https://github.com/marnusw/date-fns-tz?tab=readme-ov-file#formatintimezone

duncan-oxd (approved):
Reading the date-fns-tz documentation, I think this is correct and it will work, so I'm approving it! I put in a suggestion for using a different function if you want, but I think the result will be the same either way 👍

src/cms/src/api/emergency-alert/content-types/emergency-alert/lifecycles.js
ayumi-oxd: Thanks! Tried `formatInTimezone` but was getting an error: `Format string contains an unescaped latin alphabet character 'A'.` for some reason ...

PR #1551, opened 2024-12-12
CMS-568: Update navigation spacing on mobile

ayumi-oxd:
### Jira Ticket:
CMS-568
CMS-569

### Description:
- CMS-568: Update navigation spacing on mobile
- CMS-569: Update container width on the find a park page

### DIFF
--- a/src/gatsby/src/styles/pageContent/pageMenu.scss
+++ b/src/gatsby/src/styles/pageContent/pageMenu.scss
@@ -80,7 +80,7 @@
       width: 100%;
       height: 68px !important;
       border: none;
-      padding: 0 32px 0 20px;
+      padding: 0 40px 0 20px;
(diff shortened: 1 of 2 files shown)


src/gatsby/src/styles/pageContent/pageMenu.scss
```diff
@@ -80,7 +80,7 @@
       width: 100%;
       height: 68px !important;
       border: none;
-      padding: 0px 20px;
+      padding: 0 32px 0 20px;
```
duncan-oxd, line 83: This works, but I noticed that the edge of the icon is right up against the edge of the text cutoff with this padding:
image
You might consider increasing it to 40px or so, to add a bit of space:

image

duncan-oxd (approved):

PR #1552, opened 2024-12-13
CMS-546: Add snowplow custom calls

ayumi-oxd:
### Jira Ticket:
CMS-546

### Description:
- Add snowplow custom calls to track
  - which filter is clicked on the Find a Park page
  - which park name is selected
  - which city name is selected
  - which keyword is used to search parks
  - which accordion is clicked on the Park page
  - which megamenu link/footer link is clicked
- For testing: We can't really test it on the local and DEV, so we need to check it on Alpha-TEST when it's merged

### DIFF
--- a/src/gatsby/src/components/footer.js
+++ b/src/gatsby/src/components/footer.js
@@ -2,9 +2,10 @@ import React from "react"
 import { graphql, useStaticQuery } from "gatsby"
 import { StaticImage } from "gatsby-plugin-image"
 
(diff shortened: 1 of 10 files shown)


duncan-oxd (approved):
Looks good to me. I'm not familiar with Snowplow's API or their analytics dashboard, but I was asking on Slack if perhaps there's a way to passively track *all* link and button clicks, and then filter it in the dashboard?

Either way, having clear descriptions with the events here will be nice 👍

PR #1554, opened 2024-12-17
CMS-484: Accessibility updates

ayumi-oxd:
### Jira Ticket:
CMS-484
CMS-485
CMS-488
CMS-489
CMS-490
CMS-492
CMS-494
CMS-496
CMS-498
CMS-500
CMS-533
CMS-536
CMS-540
CMS-542
CMS-550
CMS-551
CMS-552
CMS-553
CMS-554
CMS-570

### Description:
- CMS-484, 485: Contain an element with role=menuitem in an element with role=menu
  - Adding role to `ul` and `li` caused a warning: "Non-interactive elements should not be assigned interactive roles jsx-a11y/no-noninteractive-element-to-interactive-role" so changed these elements to `div`
- CMS-488, CMS-494: Remove role=button from park photo gallery
  - Remove `roll=button` from `div`
  - Change the element around the image to `button` so that users can click/tab focus on each image
- CMS-489: Add aria-controls to input
  - Add `aria-control` to the park name search and the city name search input field 
  - Add `aria-control` to the filter checkbox input fields
- CMS-490, 533: Add for to advisories search filter label
- CMS-492: Add underline to links under the more ways to find a park
- CMS-496: Add title to iframe
  - Remove HTMLArea file and replace it with HtmlContent since it was a duplicate
- CMS-498: Apply aria-label to see all link on find a park page
- CMS-500: Add lang attribute to htmlContent
  - `lang="en"` is already set at `<html>` but add it again to `htmlContent` to ensure
- CMS-536: Add unique id to popular filters on find a park page
- CMS-540: Move focus when the scroll button is clicked by keyboard
  - Move focus to the last card item when the load more button is clicked by keyboard on the Active advisories page
- CMS-542: Add urgency level name to advisory icon
  - Update `aria-label` based on urgency level
- CMS-550: Change heading on find a park page
  - Change `h3` to `p` since the page doesn't have `h1` and `h2` 
- CMS-551: Remove alt from fontawesome icon
- CMS-552: Redirect to the park page if there is one result that exactly matches input text
- CMS-553: Add aria label to show all links and expand all links
  - Update show all links and expand all links name
- CMS-554: Add aria live to no match message in typeahead fields
- CMS-570: Change aria label on go back to the full list button
- Remove unused component files


src/gatsby/src/components/park/parkOverview.js
```diff
@@ -51,15 +51,10 @@ export default function ParkOverview({ data: parkOverview, type }) {
       {hasExpandCondition &&
         <button
           className="btn btn-link park-overview-link expand-icon"
+          aria-label={expanded ? "Show less highlights" : "Show more highlights"}
```
duncan-oxd, line 54: I think this came up as a controversy late last year. Do you remember if a decision was made about "show less" vs. "show fewer?"

ayumi-oxd: We decided to go with just "Show less/more" (https://bcparksdigital.atlassian.net/browse/CMS-441), but it'd be better to have a more specific label for the VoiceOver users.

duncan-oxd: In that case, I think we should update it to be gramatically correct. Same with the 3 aria-labels in desktopFilters.js: "Show _fewer_ areas", etc.

src/gatsby/src/pages/find-a-park.js
```diff
@@ -720,6 +724,13 @@ export default function FindAPark({ location, data }) {
     }
   }, [currentLocation])
 
+  // redirect to the park page if there is only one search result that exactly matches the input text
+  useEffect(() => {
+    if (searchResults.length === 1 && searchResults[0].protectedAreaName.toLowerCase() === inputText.toLowerCase()) {
+      navigate(`/${searchResults[0].slug}`)
+    }
+  }, [searchResults, inputText])
```
duncan-oxd, line 732: Will this happen instantly? Should we wait for an "enter" key press or something?

ayumi-oxd: You can search and redirect to the park page either way:
- Type something (e.g. "Adam") and select the specific park (e.g. "Adams Lake Marine Park") from the dropdown
- Type the exact park name (e.g.  "Adams Lake Marine Park") and hit the enter key or the search button

I think the first option is a more common use case. In that case, it happens instantly.

duncan-oxd: I thought the useEffect would make this happen every time you typed something to change the inputText, but I see there's a separate searchText variable and we're checking for the Enter key elsewhere 👍

duncan-oxd (approved):

PR #1559, opened 2025-01-06
CMS-34: Add park name audio collection

ayumi-oxd:
### Jira Ticket:
CMS-34
CMS-182

### Description:
- Add the audio clip collection to Strapi
- Users can listen to audio in the TLDR section and the Highlights section on the Park/Site page
- Users can see audio credit and audio transcript
- Editors can add an audio file to the TLDR if they choose `audioClipType` "Park name"
- Editors can add an audio file to the Highlights if they choose `audioClipType` NOT "Park name" (TBD)
- Editors can inject an audio file into the placeholder position in `description` in the protected area collection
---
#### Additional description:
- Add `displayLocation` that allows users to choose multiple options
- Install `strapi-plugin-multi-select` https://www.npmjs.com/package/strapi-plugin-multi-select to add `displayLocation`
- Update the audio button colour due to the design update
- Add `adminNote`


src/gatsby/src/styles/audioButton.scss
```diff
@@ -0,0 +1,157 @@
...
+  position: fixed;
+  bottom: 0;
+  left: 0;
+  z-index: 1025;
+  padding: 32px;
+  @media (min-width: $lgBreakpoint) {
+    height: 125px;
+  }
+  &-container {
+    width: 1100px;
```
duncan-oxd, line 33: Will that work on phones? It looks like we might need to use a breakpoint for smaller screens, but as of yet I haven't run this locally to see how it looks

src/gatsby/src/styles/audioButton.scss
```diff
@@ -0,0 +1,157 @@
...
+  width: 100%;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  color: $colorWhite;
+  background-color: $colorGrey;
+  position: fixed;
+  bottom: 0;
+  left: 0;
+  z-index: 1025;
```
duncan-oxd, line 27: It would be helpful to have a comment saying why we're using this specific number

src/gatsby/src/styles/audioButton.scss
ayumi-oxd: It works on my end unless I shrink the screen size to less than 300 px or something but let me know if it doesn't. If you're interested in UI on mobile without running it locally, please take a look at the Figma 🙂

src/gatsby/src/styles/audioButton.scss
ayumi-oxd: Good catch! Added some comments.

duncan-oxd (approved):
Very brave to create our own component instead of adapting a pre-built one!

src/gatsby/src/components/audioButton.js
```diff
@@ -0,0 +1,163 @@
...
+
+  // functions
+  const stripHtmlTags = html => {
+    const div = document.createElement("div")
+    div.innerHTML = html
+    return div.textContent || div.innerText || ""
+  }
+  const createVttContent = transcript => {
+    const plainTextTranscript = stripHtmlTags(transcript)
+    return `WEBVTT\n\n1\n00:00:00.000 --> 00:00:10.000\n${plainTextTranscript}`
```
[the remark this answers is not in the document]
ayumi-oxd: Good catch! I forgot to update it 😅

duncan-oxd (approved):

PR #1560, opened 2025-01-08
CMS-545: Update api fetching on park operating dates page

ayumi-oxd:
### Jira Ticket:
CMS-545

### Description:
- Update API fetching on the park operating dates page
- Previously the API was called 1000 times as the same as the number of parks. 
After talking to @fatbird we've decided to change the API call rather than changing the API rate limit:
  - Fetch all the advisories so that we just need one API call
  - Filter advisories with `protectedArea.orc` so that we can determine each park access status


duncan-oxd (approved):
Batching requests whenever possible is a good idea! We might still want to change the rate limit on the dev environment, because I often have trouble even building the site with the Dev Strapi because of how many calls it does. But this looks like a big improvement for the public site 👍

PR #1564, opened 2025-01-16
CMS-542: Accessibility updates and fixes

ayumi-oxd:
### Jira Ticket:
CMS-485
CMS-542
CMS-548
CMS-549
CMS-550
CMS-554
CMS-643

### Description:
- CMS-485: Add menuitem role to megamenu back button
- CMS-542: Add separation between urgency icon and advisory title
- CMS-548: Skip reading urgency icon label on active advisories page
- CMS-549: Add visually hidden link type to contact link
  - It depends on how you guide the VO, but the VO could read only static text, not `aria-label`. Add visually hidden static text in case the VO doesn't read `aria-label`
- CMS-550: Adjust heading level on list pages
- CMS-554: Update status message in typeahead fields
  - VO reads the status message if there's no match, but it doesn't read full text "No match. Please check your spelling." 🤔
- CMS-643: Add aria-hidden to expandable html content
  - VoiceOver reads the content after users click the "Show more" button

### DIFF
--- a/src/gatsby/src/components/park/contact.js
+++ b/src/gatsby/src/components/park/contact.js
@@ -61,7 +61,7 @@ export const ParkContact = ({ contact }) => {
                 href={`${getPrefix(link.contactType)}${link.contactUrl}`}
                 aria-label={`${link.contactType}: ${link.contactText}`}
               >
-                {link.contactText}
+                <span className="visually-hidden">{link.contactType}: </span>{link.contactText}
               </a>
               :
               link.contactText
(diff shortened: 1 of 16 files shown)


src/gatsby/src/components/park/contact.js
```diff
@@ -108,11 +108,15 @@ export default function Contact({ contact, parkContacts, operations }) {
                     </p>
                     <p>
                       <FontAwesome icon="phone" />
-                      <a href="tel:+18006899025" aria-label="Phone number: 1-800-689-9025">1-800-689-9025</a> (toll-free from Canada or the US)
+                      <a href="tel:+18006899025" aria-label="Phone number: 1-800-689-9025">
+                        <span className="visually-hidden">Phone number: </span>1-800-689-9025
+                      </a> (toll-free from Canada or the US)
```
duncan-oxd, line 113: Is the aria-label still necessary if the inner text of the tag says the same thing now? (I'm not sure how that works. Will screen readers read it twice?).

duncan-oxd (approved):
Looks good! I had a question about those blocks with aria labels that match their content, but you can merge if that's not an issue 👍

src/gatsby/src/components/park/contact.js
ayumi-oxd: Good catch! It doesn't read twice because `aria-label` is usually prioritized and ignores elements inside `<a>`, but I agree we don't have to have it since we have the inner text and it works for most cases.

PR #1570, opened 2025-01-23
CMS-485: Replace nav with div not to contain elements with other roles

ayumi-oxd:
### Jira Ticket:
CMS-485

### Description:
- Replace `<nav>` with `<div>` not to contain elements with other roles inside the element with `role=menu`/`role=menuitem`
- SortSite warning:
```
Elements with role=menu must contain or own an element
with role=menuitem or role=menuitemcheckbox or role=menuitemradio
and must not contain elements with other roles. Owned roles: navigation
```

### DIFF
--- a/src/gatsby/src/components/megaMenu.js
+++ b/src/gatsby/src/components/megaMenu.js
@@ -247,7 +247,7 @@ const MegaMenu = ({ content, menuMode }) => {
     return (
       <>
         {item.hasChildren && (
-          <nav
+          <div
             className={
               "menu-level menu-level--" + item.treeLevel +
               " has-clicked-twice--" + hasClickedTwice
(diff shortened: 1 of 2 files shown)


duncan-oxd (approved):
It looks like everything already was a "menuitem" but maybe it had a problem with some of the container elements etc. in the megamenu? 

We still have a div with role=menu in there so it still looks good to me 👍

ayumi-oxd:
> It looks like everything already was a "menuitem" but maybe it had a problem with some of the container elements etc. in the megamenu?
> 
> We still have a div with role=menu in there so it still looks good to me 👍

`generateMenus` is recursive and it looks like that made this error, but I can't use SortSite locally so I'll see if this could fix it once it's merged🤞

PR #1572, opened 2025-02-03
CMS-678: Add featureId to park operation sub area

ayumi-oxd:
### Jira Ticket:
CMS-678

### Description:
- Add a new integer field `featureId` to park operation sub area as per DOOT's request

### DIFF
--- a/src/cms/src/api/park-operation-sub-area/content-types/park-operation-sub-area/schema.json
+++ b/src/cms/src/api/park-operation-sub-area/content-types/park-operation-sub-area/schema.json
@@ -12,6 +12,10 @@
     "draftAndPublish": true
   },
   "attributes": {
+    "featureId": {
+      "type": "integer",
+      "unique": true
+    },
     "parkSubArea": {
       "type": "string"
     },


src/cms/src/api/park-operation-sub-area/content-types/park-operation-sub-area/schema.json
```diff
@@ -12,6 +12,10 @@
     "draftAndPublish": true
   },
   "attributes": {
+    "featureId": {
+      "type": "integer",
+      "unique": true
+    },
```
diego-oxd, line 18: Can we make it a string? 
The idea was to use `{orcs}_{prodSubareaId}`. Or open to change the script if you have an idea of something else we can use for this ID,

diego-oxd (approved):

src/cms/src/api/park-operation-sub-area/content-types/park-operation-sub-area/schema.json
ayumi-oxd: Yes we can change it to string. BTW what would be `prodSubareaId`?

diego-oxd (approved):

src/cms/src/api/park-operation-sub-area/content-types/park-operation-sub-area/schema.json
diego-oxd: The ID of the equivalent park-operation-sub-area in the prod environment would be used as part of this featureId field in all envs.

PR #1573, opened 2025-02-07
CMS-686: Update snowplow events

ayumi-oxd:
### Jira Ticket:
CMS-686

### Description:
- Change `label` to null for `search` action
- Add `url` to snowplow event obj for `link_click` action
- Remove “… accordion“ from `label` for `accordion_open` and `accordion_close` actions
- Fire the snowplow event with accumulated search conditions
  - Not to clear the conditions when filters are selected/deleted
  - Use useState to accumulate the conditions
- Fix incorrect search result number

### DIFF
--- a/src/gatsby/src/pages/find-a-park.js
+++ b/src/gatsby/src/pages/find-a-park.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useCallback, useMemo } from "react"
+import React, { useState, useEffect, useCallback, useMemo, useRef } from "react"
 import { graphql, Link as GatsbyLink } from "gatsby"
 import axios from "axios"
 import { orderBy } from "lodash"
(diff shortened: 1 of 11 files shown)


src/gatsby/src/pages/find-a-park.js
```diff
@@ -719,6 +707,55 @@ export default function FindAPark({ location, data }) {
...
 
+  // calc park results count to display
+  useEffect(() => {
+    const value = 
+      selectedCity.length > 0 &&
+      (selectedCity[0].latitude !== 0 && selectedCity[0].longitude !== 0) &&
+      hasParksWithinFifty(searchResults) ? 
+        totalResultsWithinFifty : totalResults
+    setFinalResults(value)
+  }, [selectedCity, searchResults, totalResultsWithinFifty, totalResults])
```
duncan-oxd, line 718: If you're using a useEffect hook just to set a variable, it might be better to use useMemo. Then you don't need the separate "set" method. Something to consider in this file for finalResults and eventParams.

duncan-oxd: (Or maybe the useEffect block for eventParams is necessary because it's only updating under certain conditions)

ayumi-oxd: Good catch! Actually, that makes sense to use useMemo for the `value` part 🙂

PR #1579, opened 2025-02-17
CMS-667: Add a relation with site to park-operation-sub-area

ayumi-oxd:
### Jira Ticket:
CMS-667

### Description:
- Add a relation with `site` to `park-operation-sub-area`
- TODO: Remove `orcsSiteNumber` from `park-operation-sub-area` after the data migration
- ^ Does this affect to DOOT the data model?


duncan-oxd (approved):
👍 Confirmed on the call this morning that changes to the "Site" model won't affect DOOT. Thanks for checking!

PR #1580, opened 2025-02-18
CMS-643: Fix an issue that users can't focus on the highlights section when there is no expand toggle

ayumi-oxd:
### Jira Ticket:
CMS-643

### Description:
- Fix an issue that users can't focus on the highlights section when there is no expand toggle
- The second issue that is mentioned in the ticket is not feasible. This is because the content is hidden with CSS, yet there is complete HTML content. Therefore, we can’t select just the visible part to let VO read. 
> Also, even if there is a link, we should first read what is already visible and then navigate to the expand link to expand and access the rest of the hidden content.

### DIFF
--- a/src/gatsby/src/components/park/campingDetails.js
+++ b/src/gatsby/src/components/park/campingDetails.js
@@ -74,7 +74,7 @@ export const CampingType = ({ camping, parkOperation }) => {
           </h3>
         </div>
         {/* let voiceover skip reading content if it is not expanded */}
-        <HtmlContent ariaHidden={hasExpandCondition ? !expanded : false} className="park-camping-description">
+        <HtmlContent ariaHidden={hasExpandCondition && !expanded} className="park-camping-description">
           {hasExpandCondition ? (expanded ? campingDescription : collapsedDescription) : campingDescription}
         </HtmlContent>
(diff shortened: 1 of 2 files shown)


src/gatsby/src/components/park/campingDetails.js
```diff
@@ -73,8 +73,9 @@ export const CampingType = ({ camping, parkOperation }) => {
             {camping?.campingType?.campingTypeName}
           </h3>
         </div>
-        <HtmlContent ariaHidden={!expanded} className="park-camping-description">
-          {expanded ? campingDescription : collapsedDescription}
+        {/* let voiceover skip reading content if it is not expanded */}
+        <HtmlContent ariaHidden={hasExpandCondition ? !expanded : false} className="park-camping-description">
```
duncan-oxd, line 77: Would this make sense to simplify these ternaries?

```suggestion
        <HtmlContent ariaHidden={hasExpandCondition && !expanded} className="park-camping-description">
```

ayumi-oxd: Great, yes that makes sense. Thanks!

duncan-oxd (approved):

PR #1584, opened 2025-02-20
CMS-672: Update advisory sorting

ayumi-oxd:
### Jira Ticket:
CMS-672

### Description:
- Update advisory sorting
- New sorting order
1.   listingRank:DES
2.   urgency.sequence:DESC
3.   accessStatus.precedence:ASC
4.   eventType.precedence:ASC
5.   Display date


duncan-oxd (approved):
Makes sense, but I'd suggest leveraging a pre-made sorting function like lodash/oderBy which could really simplify the code here.
If the keys like `listingRank` and `urgency.sequence` are potentially undefined then it might not clean it up _that_ much... but I'd still recommend it to avoid potential typos and keep the code concise

PR #1589, opened 2025-03-05
CMS-755: Navigate to park page from home page if park is selected

ayumi-oxd:
### Jira Ticket:
CMS-755

### Description:
- Navigate to the park page from the home page if the park is selected from the option
  - It directly goes to the Golden Ears Park page if a user selects "Golden Ears Park" from the option
  - It goes to the Find a park page if a user searches "Gold"

### DIFF
--- a/src/gatsby/src/components/search/mainSearch.js
+++ b/src/gatsby/src/components/search/mainSearch.js
@@ -35,6 +35,7 @@ const MainSearch = ({ hasCityNameSearch }) => {
   const [selectedCity, setSelectedCity] = useState([])
   const [hasPermission, setHasPermission] = useState(false)
   const [isMatched, setIsMatched] = useState(false)
+  const [hasParkSelected, setHasParkSelected] = useState(false)
   const [currentLocation, setCurrentLocation] = useState({
     strapi_id: 0,
     cityName: "Current location",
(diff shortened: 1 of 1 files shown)


src/gatsby/src/components/search/mainSearch.js
```diff
@@ -95,7 +102,8 @@ const MainSearch = ({ hasCityNameSearch }) => {
   // event handlers
   const handleSearchNameChange = (selected) => {
     if (selected.length) {
-      setSearchText(selected[0]?.protectedAreaName)
+      setSearchText(selected[0].protectedAreaName)
+      setSelectedParkSlug(selected[0].slug)
     }
```
duncan-oxd, line 107: Should there be an `else` here (or somewhere) that clears that state if the selection or search input changes before they click?

ayumi-oxd: Good point! This function `handleSearchNameChange` works only if a user selects a park from the option: https://github.com/ericgio/react-bootstrap-typeahead/blob/main/docs/API.md So `else` condition won't be triggered (then maybe I don't need the if statement ... but to make sure `selected` isn't an empty array).

If a user selects a park, it automatically goes to the respective park page, so there will be no input change. If a user comes back to the home page again from the park page, the states will be cleared. I thought adding `setSelectedParkSlug("")` in the `handleClickClearPark`, but I assume there'll be no situation like that because of this redirecting.

If a user normally enters a keyword and clicks search, `handleSearchNameInputChange` works but not `handleSearchNameChange`.

duncan-oxd: Thanks for the explanation 👍  As long as there's no case where the `selectedParkSlug` is out of sync with the item they're selecting

duncan-oxd (approved):

PR #1590, opened 2025-03-07
CMS-668: Fix a broken link accessibility error

ayumi-oxd:
### Jira Ticket:
CMS-123

### Description:
- Fix a broken link accessibility error
- Change it to hide the link on mobile with the screen size state condition, not with CSS
- Error:
> A skip navigation link exists, but the target for the link does not exist or the link is not keyboard accessible.
> Ensure that the target for the link exists and that the link is not hidden with CSS display:none or visibility:hidden.

### DIFF
--- a/src/gatsby/src/components/header.js
+++ b/src/gatsby/src/components/header.js
@@ -3,12 +3,13 @@ import React from "react"
 import SkipToContent from "./skipToContent"
 import MegaMenu from "./megaMenu.js"
 import EmergencyAlert from "./emergencyAlert"
+import { useScreenSize } from "../utils/helpers"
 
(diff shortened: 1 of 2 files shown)


duncan-oxd (approved):
Hm! I will have to read about best practices for these kinds of "Skip to content" links. I didn't realize hiding them with CSS was an issue

ayumi-oxd:
> Hm! I will have to read about best practices for these kinds of "Skip to content" links. I didn't realize hiding them with CSS was an issue

Another way is changing its position with CSS, but I choose not to display the element itself because I don't want to be bothered again by the accessibility tools 😅

PR #1591, opened 2025-03-07
CMS-767: Update snowplow track event filters

ayumi-oxd:
### Jira Ticket:
CMS-767

### Description:
- Update snowplow track event filters
- Use lowercase for all filter names
- Rename "Things to do" to "activities" to align with the schema


duncan-oxd (approved):
Looks good! It might be good to get in the habit of using `toLocaleLowerCase` in place of `toLowerCase` most of the time, but if the values are known and it won't make a difference, either one works 👍 
(Just thinking about lowercasing park names with non-ASCII characters)

PR #1593, opened 2025-03-10
CMS-256: Add land acknowledgment

ayumi-oxd:
### Jira Ticket:
CMS-256

### Description:
- Add land acknowledgment

### DIFF
--- a/src/gatsby/src/components/footer.js
+++ b/src/gatsby/src/components/footer.js
@@ -115,6 +115,13 @@ export default function Footer() {
 
   return (
     <footer id="footer">
+      <div className="acknowledgment-container">
+        <div className="acknowledgment">
+          We acknowledge all First Nations on whose territories BC Parks were established. 
+          We honour their connection to the land and respect the importance of their diverse teachings,
+          traditions, and practices within these territories.
+        </div>
+      </div>
       <div className="home-footer" id="home-footer">
         <div className="row g-0">
           <div className="col col-12 col-md-4">
--- a/src/gatsby/src/styles/footer.scss
+++ b/src/gatsby/src/styles/footer.scss
@@ -4,17 +4,47 @@
 @import "~bootstrap/scss/mixins";
 
 #footer {
+  position: relative;
   background-color: $colorBlue !important;
-  overflow-x: hidden;
+}
+
+.acknowledgment-container {
+  position: absolute;
+  top: -100px;
+  left: 16px;
+  height: 200px;
+  width: calc(100% - 32px);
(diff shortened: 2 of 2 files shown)


src/gatsby/src/components/acknowledgment.js
```diff
@@ -0,0 +1,17 @@
+import React from "react"
+import "../styles/acknowledgment.scss"
+
+export default function Acknowledgment({ color, condition }) {
+  return (
+    <div className={`acknowledgment is-bg-${color}--${condition}`}>
```
duncan-oxd, line 6: I'd suggest using a more descriptive value than "true" here - or at least putting some comments in to make it clear what that prop does. I see it being passed in as a boolean, but then we're only using the string value? How about using another class that we add if the prop is true?

src/gatsby/src/styles/staticLanding1.scss
```diff
@@ -62,10 +62,10 @@
 .bcp-landing-links {
   width: 100%;
   background-color: $colorBackgroundGrey;
-  padding: 32px 0;
+  padding: 32px 0 0;
```
duncan-oxd, line 65: It's usually easier to maintain styles if you do spacing on the bottom of the element above, instead of adding top margins (or at least avoid mixing the two whenever you can help it). I thought I would mention because this is a small stylesheet but it might be easier to leave it the way it is, since this site has an established codebase

duncan-oxd:
> TODO: It might be better to create a component like that includes Header, Acknowledgment, ScrollToTop, and Footer. Then it can wrap content as a prop.

I agree 👍  In a future update, we can make a layout component that wraps the content as a child component (react router calls it an "Outlet")

duncan-oxd (changes requested):
One small note to clarify the "condition" prop, but the changes look good otherwise!

src/gatsby/src/components/acknowledgment.js
ayumi-oxd: I updated it to just a prop `color` and the class will be added based on `color`. The bottom of the park page can have a background color depending on `nearByParks` component, so one condition is used when passing `color` prop in `Acknowledgement`.

duncan-oxd: Looks good. You might find this package handy, we're using it on DOOT:
https://www.npmjs.com/package/classnames
It makes it a little easier to add conditional class names with object/array forms. (It's how classes work in Vue JS)

duncan-oxd (approved):

PR #1594, opened 2025-03-12
CMS-777: Fix keycloak login issue

ayumi-oxd:
### Jira Ticket:
CMS-777

### Description:
- Fix an issue that the users couldn't log in to the Staff Portal by updating `keycloak-js`
- This is the same as this fix: https://github.com/bcgov/bcparks-staff-portal/pull/144, but for PROD since they will make a change on March 25th
- If DOOT will be deployed on the OpenShift Silver PROD before 25th, this is not needed, but it doesn't harm the admin app even if it's merged

### DIFF
--- a/src/admin/package-lock.json
+++ b/src/admin/package-lock.json
@@ -25,7 +25,7 @@
         "bootstrap": "^4.6.0",
         "history": "^5.0.0",
         "jsonwebtoken": "^8.5.1",
-        "keycloak-js": "^12.0.4",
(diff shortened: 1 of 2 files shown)


duncan-oxd (approved):

ayumi-oxd:
Close this PR as the Staff Portal from https://github.com/bcgov/bcparks-staff-portal is deployed on PROD due to the DNS change.

PR #1595, opened 2025-03-12
CMS-771: Update bcparks map link

ayumi-oxd:
### Jira Ticket:
CMS-771

### Description:
- Update BC parks map link
- Add a constant of the link so it can be used any files


duncan-oxd (approved):
Good idea making it a constant 👍

PR #1596, opened 2025-03-13
CMS-776: Change wording winter season to winter rate

ayumi-oxd:
### Jira Ticket:
CMS-776

### Description:
- Change wording winter season to winter rate
- Add a tooltip next to "Winter rate" on the park page
- Remove "Winter season" column on the park operating dates page
- If the winter rate dates are not available "Winter rate" will not be displayed

### DIFF
--- a/src/gatsby/src/styles/parks.scss
+++ b/src/gatsby/src/styles/parks.scss
@@ -563,8 +563,9 @@ h2.section-heading {
       border-radius: 8px;
       margin: 0;
       padding: 0;
+      // change the color of the icon to $colorGrey
       img {
-        filter: brightness(0) saturate(100%) invert(7%) sepia(0%) saturate(0%) hue-rotate(253deg) brightness(97%) contrast(97%);
+        filter: invert(7%) brightness(20%);
       }
     }
   }
--- a/src/gatsby/src/styles/parks.scss
+++ b/src/gatsby/src/styles/parks.scss
@@ -556,6 +556,17 @@ h2.section-heading {
         height: 16px;
       }
     }
+    .btn-tooltip {
+      width: 16px;
+      height: 16px;
+      display: inline-flex;
+      border-radius: 8px;
+      margin: 0;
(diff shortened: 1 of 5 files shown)


src/gatsby/src/styles/parks.scss
```diff
@@ -556,6 +556,17 @@ h2.section-heading {
...
     }
+    .btn-tooltip {
+      width: 16px;
+      height: 16px;
+      display: inline-flex;
+      border-radius: 8px;
+      margin: 0;
+      padding: 0;
+      img {
+        filter: brightness(0) saturate(100%) invert(7%) sepia(0%) saturate(0%) hue-rotate(253deg) brightness(97%) contrast(97%);
```
duncan-oxd, line 567: What is this doing? Let's add a comment to clarify the purpose.
I'm also guessing it can be simplified... if brightness is 0, won't it just appear black, and then the other filters can be removed?

duncan-oxd (approved):
Changes look good! I had one question about the CSS. If we add a comment, you can go ahead and merge after that 👍

src/gatsby/src/styles/parks.scss
ayumi-oxd: This is to change the colour of the icon image to #1a1a1a. I guess it can be simplified like `invert(7%) brightness(97%)`?

duncan-oxd: Oh! I didn't notice the second brightness function in there, so I thought the brightness 0 just made it black :D 
It looks like it definitely can be simplified by taking out the rules that don't have any effect in this case 👍

duncan-oxd: One thing to note: if the images are SVGs, you can edit he SVG file to use `currentColor` instead of a hardcoded value like `#FF0000` and then it will use whatever the css color of the container is. That's how FontAwesome icons can change color. It might be useful on this site if we're displaying icons in different colors.

PR #1598, opened 2025-03-18
CMS-695: Display sub area dates from park feature date collection

ayumi-oxd:
### Jira Ticket:
CMS-695

### Description:
- Get sub area dates from `parkFeatureDate` and `parkOperationSubAreaDate` collections through Strapi API, not GraphQL so that editors can update the dates without the Gatsby build
- Display sub area dates on the Park, Site, and Park operating dates page
- Prioritize dates from `parkFeatureDate` to display, if it doesn't exist display dates from `parkOperationSubAreaDate`
- TODO: Delete `parkOperationSubAreaDate` once we finish data migration from `parkOperationSubAreaDate` to `parkFeatureDate`
- TODO: Test it by publishing dates on DOOT ALPHA-DEV and check if the date is created/updated on Strapi/FE

### DIFF
--- a/src/gatsby/src/templates/park.js
+++ b/src/gatsby/src/templates/park.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useState, useRef } from "react"
+import React, { useEffect, useState, useRef, useMemo } from "react"
 import axios from "axios"
 import { sortBy, truncate } from "lodash"
 import { graphql, Link as GatsbyLink, navigate } from "gatsby"
@@ -84,11 +84,8 @@ export default function ParkTemplate({ data }) {
   const [parkAccessStatus, setParkAccessStatus] = useState(null)
   const [addedSeasonalAdvisory, setAddedSeasonalAdvisory] = useState(false)
   const [subAreas, setSubAreas] = useState([])
-  const [processSubAreas, setProcessSubAreas] = useState([])
   const [subAreasLoadError, setSubAreasLoadError] = useState(false)
   const [isLoadingSubAreas, setIsLoadingSubAreas] = useState(true)
-  const activeFacilities = combineFacilities(park.parkFacilities, data.allStrapiFacilityType.nodes, subAreas)
-  const activeCampings = combineCampingTypes(park.parkCampingTypes, data.allStrapiCampingType.nodes, subAreas)
 
   const loadAdvisoriesData = async () => {
     setIsLoadingAdvisories(true)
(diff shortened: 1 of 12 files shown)


src/gatsby/src/templates/park.js
```diff
@@ -87,40 +83,91 @@ export default function ParkTemplate({ data }) {
...
+    setIsLoadingSubAreas(true)
+    try {
+      const response = await loadSubAreas(apiBaseUrl, park.orcs)
+      if (response.status === 200) {
+        setSubAreas(response.data.data)
+        setSubAreasLoadError(false)
+      } else {
+        setSubAreas([])
+        setSubAreasLoadError(true)
+      }
```
duncan-oxd, line 142: Is there ever a case where the API would send a successful response other than 200, and we wouldn't want to use the response data? I wonder if we even need the check for the status and the `else` bock, because I'd expect axios to throw an exception if the response failed. Just curious! I haven't read about the Strapi public API yet.

duncan-oxd (approved):

src/gatsby/src/templates/park.js
ayumi-oxd: Good point! Since I changed it to try and catch, I can simplify `response.status ...` part.

diego-oxd (commented):
Looks good. Just that one comment about allowing the save incomplete dates as draft. 
Also, I don't know if they wanted it to work that way. But if you open a season that is "approved" or "on API" you see the validation error for notes right away.

ayumi-oxd:
FYI @duncan-oxd I updated a function using lodash, thanks for your advice!

src/gatsby/src/utils/parkDatesHelper.js
```diff
@@ -83,61 +83,35 @@ const processDateRanges = (arr, fmt, yr, delimiter, yearPrefix) => {
...
+  // override subAreaDates with featureDates
+  const dateTypes = {
+    "Operation": "serviceDates",
+    "Reservation": "resDates",
+    "Winter fee": "offSeasonDates",
+    // TODO: add more date types as needed
   }
-  return subArea;
+  _.forEach(dateTypes, (key, type) => {
+    const featureDate = featureDates.find((date) => date.dateType === type)
```
duncan-oxd, line 109: This is still potentially looping through the `featureDates` array several times when we could do the grouping with just one loop, but the array will never be huge so it's fine either way. 
I was thinking (if `featureDates` has just one date of each type, then you can use `_.keyBy` to get an object of all the different date types, and use something like `pick` to narrow it down to just the ones you care about  (the 3 types listed in `dateTypes`, in this case)

Let me know if you want to chat about my idea, but it also looks good to merge!

ayumi-oxd: Thanks for checking, updated it!

PR #1604, opened 2025-04-09
CMS-672: Fix display date sorting for advisories

duncan-oxd:
### Jira Ticket:
CMS-672

### Description:
A small fix for the advisories sort order: We're using lodash `orderBy` here so we don't need to pass in a sort function, just an identity function to return the values to compare. I changed `compareAdvisories` to `getAdvisoryDate` to get the display date value, and `orderBy` does the comparison automatically.

Additionally, if the identity function just returns a property and doesn't do anything special, we can just supply the name as a string instead of having to write the function ourselves. I was able to simplify the other 4 sort parameters, so we've got 4 strings and just one function.

### DIFF
--- a/src/gatsby/src/components/park/advisoryDetails.js
+++ b/src/gatsby/src/components/park/advisoryDetails.js
@@ -14,7 +14,7 @@ import AdvisoryDate from "../advisories/advisoryDate"
 import blueAlertIcon from "../../images/park/blue-alert.svg"
 import redAlertIcon from "../../images/park/red-alert.svg"
 import yellowAlertIcon from "../../images/park/yellow-alert.svg"
-import { compareAdvisories } from "../../utils/advisoryHelper"
+import { getAdvisoryDate } from "../../utils/advisoryHelper"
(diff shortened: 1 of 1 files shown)


src/gatsby/src/components/park/advisoryDetails.js
```diff
@@ -191,7 +185,7 @@ export default function AdvisoryDetails({ advisories, parkType, parkAccessStatus
                         advisoryDate={advisory.formattedAdvisoryDate}
                         hasUpdatedDateDisplay={advisory.isUpdatedDateDisplayed && advisory.formattedUpdatedDate}
                         updatedDate={advisory.formattedUpdatedDate}
-                        hasDisplayedDate={advisory.isAdvisoryDateDisplayed || advisory.isUpdatedDateDisplayed || 
+                        hasDisplayedDate={advisory.isAdvisoryDateDisplayed || advisory.isUpdatedDateDisplayed ||
```
duncan-oxd, line 188: My editor did this 😓 I could undo it but I'd sooner leave it in!

ayumi-oxd (approved):

PR #1606, opened 2025-04-15
CMS-747: Fix build error

ayumi-oxd:
### Jira Ticket:
CMS-747

### Description:
- Revert `better-sqlite3` update: https://github.com/bcgov/bcparks.ca/pull/1588
- Fix CMS build error: https://github.com/bcgov/bcparks.ca/actions/runs/14460053601/job/40551037699
```
npm ERR! code 1
npm ERR! path /opt/app-root/src/node_modules/better-sqlite3
npm ERR! command failed
npm ERR! command sh -c -- prebuild-install || node-gyp rebuild --release
npm ERR! make: Entering directory '/opt/app-root/src/node_modules/better-sqlite3/build'
npm ERR!   TOUCH ba23eeee118cd63e16015df367567cb043fed872.intermediate
npm ERR!   ACTION deps_sqlite3_gyp_locate_sqlite3_target_copy_builtin_sqlite3 ba23eeee118cd63e16015df367567cb043fed872.intermediate
npm ERR!   TOUCH Release/obj.target/deps/locate_sqlite3.stamp
npm ERR!   CC(target) Release/obj.target/sqlite3/gen/sqlite3/sqlite3.o
npm ERR!   AR(target) Release/obj.target/deps/sqlite3.a
npm ERR!   COPY Release/sqlite3.a
npm ERR!   CXX(target) Release/obj.target/better_sqlite3/src/better_sqlite3.o
npm ERR! rm ba23eeee118cd63e16015df367567cb043fed872.intermediate
npm ERR! make: Leaving directory '/opt/app-root/src/node_modules/better-sqlite3/build'
npm ERR! prebuild-install warn install /lib[64](https://github.com/bcgov/bcparks.ca/actions/runs/14460053601/job/40551037699#step:5:65)/libm.so.6: version `GLIBC_2.29' not found (required by /opt/app-root/src/node_modules/better-sqlite3/build/Release/better_sqlite3.node)
```

### DIFF
--- a/src/cms/package.json
+++ b/src/cms/package.json
@@ -37,7 +37,7 @@
     "npm": ">=6.0.0"
   },
   "devDependencies": {
-    "better-sqlite3": "^11.8.1",
+    "better-sqlite3": "^11.0.0",
     "jest": "^29.7.0",
     "koa": ">=2.15.4"
   },
--- a/src/cms/package-lock.json
+++ b/src/cms/package-lock.json
@@ -28,7 +28,7 @@
         "styled-components": "^5.3.11"
       },
       "devDependencies": {
-        "better-sqlite3": "^11.8.1",
+        "better-sqlite3": "^11.0.0",
         "jest": "^29.7.0",
         "koa": ">=2.15.4"
       },


src/cms/package.json
```diff
@@ -37,7 +37,7 @@
     "npm": ">=6.0.0"
   },
   "devDependencies": {
-    "better-sqlite3": "^11.8.1",
+    "better-sqlite3": "^11.0.0",
```
duncan-oxd, line 40: Is this only a dev dependency? I wouldn't think that would affect the builds... If it's needed for production, maybe it should be moved to the regular dependencies list?

duncan-oxd (approved):

src/cms/package.json
ayumi-oxd: I'm not familiar with it so I could be wrong here, but it looks like it's only for building when PR is merged. I couldn't run this s2i build locally so am not sure if it's going to fix the issue, but wanted to try a revert since it was the only change that I made ... I'd appreciate it if you have any suggestions 🙂

`build-dev.yaml`
```
      - name: s2i Build
        id: build_image
        uses: redhat-actions/s2i-build@v2
        with:
          path_context: "src/cms"
          builder_image: ${{ env.BUILDER_IMAGE }}
          image: ${{ secrets.OPENSHIFT_GOLD_EXTERNAL_REPOSITORY }}/${{ secrets.OPENSHIFT_GOLD_LICENSE_PLATE }}-tools/${{ env.IMAGE_NAME }}
          tags: latest ${{ env.SHORT_SHA }}
          env_vars: STRAPI_ADMIN_BACKEND_URL=
```

duncan-oxd: I'm not familiar with that s2i-build step either! When that action failed, the error said:

> version `GLIBC_2.29' not found

I'm guessing that `builder_image` ("registry.access.redhat.com/ubi8/nodejs-18:1-32") doesn't support that version of `glibc` -- We could try running it with a different builder_image (I don't have any specific suggestions, but a generic node 18 image might have it?) It sounds like it's pretty common, but the [UBI8 image only has 2.28](https://catalog.redhat.com/software/containers/ubi8/ubi-minimal/5c359a62bed8bd75a2c3fba8?container-tabs=packages)

Otherwise, if it works to just downgrade the version of sqlite then that would solve it too. (We'd just have to be careful about upgrading it in the future)

duncan-oxd: We could try just changing it to ubi9 instead of 8? Or `node:18-bullseye` - both have newer versions of glibc, from the looks of it.

I'd say let's try that first, then we can leave the updated sqlite package... and or downgrade it as a plan B if the github action still fails 🤔

ayumi-oxd:
Close it since it fixed the issue: https://github.com/bcgov/bcparks.ca/pull/1608

PR #1607, opened 2025-04-15
CMS-823: Fix displaying facilities with isActive false

ayumi-oxd:
### Jira Ticket:
CMS-823

### Description:
- Fix displaying facilities with isActive false
- Add filter with `isActive` and `isFacilityOpen`
- Update activities and camping filtering too

### DIFF
--- a/src/gatsby/src/utils/subAreaHelper.js
+++ b/src/gatsby/src/utils/subAreaHelper.js
@@ -49,11 +49,13 @@ const combineCampingTypes = (parkCampingTypes, campingTypes, subAreas) => {
 
   // add the parkCampingTypes to the common object
   for (const parkCampingType of parkCampingTypes) {
-    const campingTypeCode = parkCampingType.campingType?.campingTypeCode;
-    if (!obj[campingTypeCode]) {
-      obj[campingTypeCode] = { subAreas: [] };
+    if (parkCampingType.isActive && parkCampingType.isCampingOpen) {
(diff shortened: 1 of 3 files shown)


src/gatsby/src/utils/subAreaHelper.js
```diff
@@ -49,11 +49,13 @@ const combineCampingTypes = (parkCampingTypes, campingTypes, subAreas) => {
...
-      obj[campingTypeCode] = { subAreas: [] };
+    if (parkCampingType.isActive && parkCampingType.isCampingOpen) {
+      const campingTypeCode = parkCampingType.campingType?.campingTypeCode;
+      if (!obj[campingTypeCode]) {
+        obj[campingTypeCode] = { subAreas: [] };
+      }
+      obj[campingTypeCode] = { ...parkCampingType, ...obj[campingTypeCode] };
     }
-    obj[campingTypeCode] = { ...parkCampingType, ...obj[campingTypeCode] };
   }
```
duncan-oxd, line 59: Rather than wrapping this whole thing in an if statement and adding another level of indents, my preference would be to check if it's _not_ active or the park is _not_ open, and then `continue;` to the next iteration the loop. (Or perhaps there's a way to filter out the inactive items from the array before the loop happens?

The changes all look good, but an "early return" might cut down on indents and keep this easy to read

ayumi-oxd: Sounds good, filter looks cleaner 👍

duncan-oxd (approved):

PR #1608, opened 2025-04-16
CMS-747: Update ubi 8 to 9 in build-dev

ayumi-oxd:
### Jira Ticket:
CMS-747

### Description:
- Update ubi 8 to 9 in build-dev based on our comments here: https://github.com/bcgov/bcparks.ca/pull/1606

### DIFF
--- a/.github/workflows/build-dev.yaml
+++ b/.github/workflows/build-dev.yaml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       IMAGE_NAME: strapi-${{ github.ref_name }}
-      BUILDER_IMAGE: registry.access.redhat.com/ubi9/nodejs-18:1-32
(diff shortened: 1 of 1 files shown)


duncan-oxd (approved):
I don't know if you pulled the docker image and tried to build the app inside it to test this, but it would be easy enough to just try the github action and see if it succeeds :D

duncan-oxd (changes requested):
oh, actually it looks like it may not exist with that "1-32" tag. We'll have to update the tag (Let's just pick whatever is the current version tag they're using)

ayumi-oxd:
> oh, actually it looks like it may not exist with that "1-32" tag. We'll have to update the tag (Let's just pick whatever is the current version tag they're using) edit: try `registry.access.redhat.com/ubi9/nodejs-18:9.5`

Good catch, thanks!

duncan-oxd (approved):
Let's try it!

PR #1613, opened 2025-04-28
CMS-823: Fix facilities display issue

ayumi-oxd:
### Jira Ticket:
CMS-823

### Description:
- Fix an issue where not all facilities are displayed on the park page
  - If there was one facility, it was displayed, but if there were more than two facilities, some of them were not displayed
- It was because the function `combineFacilities` returned a result before the for loop finished, because one of the props was passed as an empty array first
- Update the facilities/camping data with `useState` and `useEffect` on the park/site page

### DIFF
--- a/src/gatsby/src/templates/park.js
+++ b/src/gatsby/src/templates/park.js
@@ -87,6 +87,8 @@ export default function ParkTemplate({ data }) {
   const [subAreas, setSubAreas] = useState([])
   const [subAreasLoadError, setSubAreasLoadError] = useState(false)
   const [isLoadingSubAreas, setIsLoadingSubAreas] = useState(true)
+  const [activeFacilities, setActiveFacilities] = useState([])
+  const [activeCampings, setActiveCampings] = useState([])
 
   const loadAdvisoriesData = async () => {
     setIsLoadingAdvisories(true)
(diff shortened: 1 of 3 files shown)


src/gatsby/src/utils/subAreaHelper.js
```diff
@@ -63,7 +63,7 @@ const combineCampingTypes = (campings, campingTypes, subAreas) => {
   // add the campingTypes to the common object and convert it to an array
   for (const campingTypeCode in obj) {
     const parkCampingType = obj[campingTypeCode];
-    parkCampingType.campingType = campingTypes.find(ct => ct.campingTypeCode === campingTypeCode);
+    parkCampingType.campingType = await campingTypes.find(ct => ct.campingTypeCode === campingTypeCode);
```
duncan-oxd, line 66: What are the data types here? It looks like `campingTypes` is an plain array of (objects?). If `.find` doesn't return a promise, then I don't think the `await` will make a difference here.

src/gatsby/src/templates/site.js
```diff
@@ -140,8 +142,51 @@ export default function SiteTemplate({ data }) {
...
-  const activeCampings = combineCampingTypes(site.parkCampingTypes, data.allStrapiCampingType.nodes, processedSubAreas);
+  // set active facilities
+  useEffect(() => {
+    const fetchActiveFacilities = async () => {
+      if (site.parkFacilities.length > 0 &&
+        data.allStrapiFacilityType.nodes.length > 0 && 
+        Object.keys(processedSubAreas).length > 0
+      ) {
+        try {
+          const facilities = await combineFacilities(
```
duncan-oxd, line 153: If the `await` in `combineFacilities` doesn't actually wait for anything, then this promise will just fulfill instantly, and in that case we could probably take out the `async`/`awaits`, unless there's something I'm missing. I'm not familiar with the data type passed to the `SiteTemplate` component here...

duncan-oxd (changes requested):
A few questions about the data types; the async code might not be necessary if it's all loaded already 🤔

ayumi-oxd:
> > It was because the function `combineFacilities` returned a result before the for loop finishes
> 
> I'm guessing the `useEffect` code is what's fixing this issue, and not the async/await, but maybe we can walk through it if I'm getting confused :D

Sorry for the confusion! I tried the async/await approach first and went to the useState/useEffect approach, but weirdly kept it both ... It should work just useState/useEffect.

src/gatsby/src/templates/park.js
```diff
@@ -149,8 +151,41 @@ export default function ParkTemplate({ data }) {
...
+      ) {
+          const facilities = combineFacilities(
+            park.parkFacilities,
+            data.allStrapiFacilityType.nodes,
+            processedSubAreas
+          )
+          setActiveFacilities(facilities)
+      }
+    }
+    fetchActiveFacilities()
```
duncan-oxd, line 169: Now that the function body doesn't have to be async, I think you can also move the code into the main useEffect body instead of defining a function and calling it right away - I think the result is the same unless the function needs to be asynchronous

duncan-oxd (approved):
Looks good to me! I added a comment about potentially saving a couple of lines and one level of indentation, but it's good to go either way 👍

PR #1615, opened 2025-04-30
CMS-34: Fix audio clip display issue

ayumi-oxd:
### Jira Ticket:
CMS-34

### Description:
- If there was no Show more/less toggle button, audio clip wasn't displayed
- If there was `<hr>` at the end of the description, the toggle button wasn't displayed
  - If there is an audio clip, display the toggle button, since it takes most of the height of the Highlight section
  - If there is `<hr>` at the end of the description, and there is an audio clip, the toggle button should be displayed

### DIFF
--- a/src/gatsby/src/components/park/parkOverview.js
+++ b/src/gatsby/src/components/park/parkOverview.js
@@ -20,7 +20,6 @@ export default function ParkOverview({ description, type, audioClips }) {
   const collapsedDescription = $.html()
   const hasHr = $("hr").length > 0
   const hrAtEnd = description.trim().endsWith("<hr>")
-  const hasExpandCondition = (hasHr || isLong) && !isMedium && !hrAtEnd
   const hasAudioClipPlaceholder = $(".audio-clip").length > 0
   // Check if array contains a "highlights"
   const hasHighlights = (array) => array?.includes("highlights") || false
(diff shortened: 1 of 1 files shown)


src/gatsby/src/components/park/parkOverview.js
```diff
@@ -20,14 +20,19 @@ export default function ParkOverview({ description, type, audioClips }) {
...
   const hasAudioClipPlaceholder = $(".audio-clip").length > 0
   // Check if array contains a "highlights"
   const hasHighlights = (array) => array?.includes("highlights") || false
   // Filter audio clips if it has a "highlights" displayLocation
   const audioClip = useMemo(() => {
     return audioClips?.filter(audio => hasHighlights(audio.displayLocation?.strapi_json_value)) || []
   }, [audioClips])
+  // Set the expand condition if 
+  // 1 - the description is long or has a <hr> tag
+  // 2 - the description is medium
```
duncan-oxd, line 32: It looks like the code is checking that `isMedium` is _false_

duncan-oxd (changes requested):
One note to check the logic (or possibly just change the comment)

src/gatsby/src/components/park/parkOverview.js
ayumi-oxd: Oops good catch!

duncan-oxd (approved):

PR #1618, opened 2025-05-05
CMS-868: Display only one audio clip at a time

ayumi-oxd:
### Jira Ticket:
CMS-868

### Description:
- Display only one audio clip (audio player) at a time
  - If a user clicks the audio button A and then clicks the audio button B, the audio player B doesn't get displayed
  - If the `activeAudio` state matches the audio id of the element, display the matched audio player

### DIFF
--- a/src/gatsby/src/components/park/parkOverview.js
+++ b/src/gatsby/src/components/park/parkOverview.js
@@ -6,7 +6,7 @@ import HtmlContent from "../htmlContent"
 import AudioButton from "../audioButton"
 import * as cheerio from "cheerio"
 
-export default function ParkOverview({ description, type, audioClips }) {
+export default function ParkOverview({ description, type, audioClips, activeAudio, setActiveAudio }) {
   const [expanded, setExpanded] = useState(false)
   const [height, setHeight] = useState(0)
   const [sectionHeight, setSectionHeight] = useState(0)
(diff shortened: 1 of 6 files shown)


src/gatsby/src/components/park/parkOverview.js
```diff
@@ -91,6 +99,7 @@ export default function ParkOverview({ description, type, audioClips }) {
         }
       })
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
duncan-oxd, line 102: Does this project have eslint configured? or do you just have it installed locally? I don't get any linter errors when I edit these files locally, and I didn't see eslint in the dependencies.

But either way, what items are missing from the dependencies array? Is there a good reason not to include them here?

ayumi-oxd: No, but I see the warning on the Chrome console.
`activeAudio` and `setActiveAudio` are not included here because I wanted to avoid unnecessary re-renders. This effect is specifically for injecting the audio element into the highlights text.

duncan-oxd: `setActiveAudio` wouldn't be changing, so that should be ok to include in the dependency array, and it looks like `activeAudio` is used to set the `audioButtonElement` variable, so wouldn't you want it to be included?

duncan-oxd: I see the update 👍 Looks good! My understanding is that we almost always want the complete list of dependencies. If a dependency is changing too much and causing the useEffect to trigger, then I think we'd look into `useMemo` or `useCallback` to make that dependency stable, rather than leaving it out of the array ✅

duncan-oxd (approved):

PR #1619, opened 2025-05-06
CMS-871: Add date format for winter rate

ayumi-oxd:
### Jira Ticket:
CMS-871

### Description:
- Add date format for winter rate
- Convert "YYYY: MM/DD – MM/DD" to "MM/DD, YYYY – MM/DD, YYYY"

### DIFF
--- a/src/gatsby/src/pages/plan-your-trip/park-operating-dates.js
+++ b/src/gatsby/src/pages/plan-your-trip/park-operating-dates.js
@@ -14,7 +14,7 @@ import ParkAccessStatus from "../../components/park/parkAccessStatus"
 import StaticIcon from "../../components/park/staticIcon"
 import NoSearchResults from "../../components/search/noSearchResults"
 import { loadAllSubAreas } from "../../utils/subAreaHelper"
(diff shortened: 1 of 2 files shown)


ayumi-oxd:
> This looks like it'll work with the dates in this unique string format. Is there somewhere earlier on that the strings are loaded as date objects or ISO strings we can parse? It would be ideal to store those values somewhere so we don't have to extract the data back out of a formatted string. But if that's the only format we have, then this will do the trick!

We use `groupSubAreaDates` to format data, but it's also used elsewhere where the dates are displayed in a different format. So I avoided making changes to `groupSubAreaDates` ... There might be a cleaner solution, but we can revisit this later since the front-end for dates will be updated soon.

duncan-oxd:
> We use `groupSubAreaDates` to format data, but it's also used elsewhere where the dates are displayed in a different format.

Sounds good! I recently made a change on DOOT so now the dates are always stored as Date objects, and they can be compared/modified/formatted any way we want in any component without having to parse the values every time. I added a response transformation function in Axios so I can show you how I did it whenever we get around to it 👍

PR #1621, opened 2025-05-09
CMS-868: Filter audio clips in about section

ayumi-oxd:
### Jira Ticket:
CMS-868

### Description:
- Previously, the same audio was used in the Heritage and the History of the About section
- Filter audio clips by the `displayLocation` and display in the matched location

### DIFF
--- a/src/gatsby/src/components/park/about.js
+++ b/src/gatsby/src/components/park/about.js
@@ -13,15 +13,15 @@ import { trackSnowplowEvent } from "../../utils/snowplowHelper"
 import "../../styles/cmsSnippets/parkInfoPage.scss"
 
 export const AccordionList = ({ eventKey, data, openAccordions, toggleAccordion, audioClips, activeAudio, setActiveAudio }) => {
-  // Check if array contains a "heritage" or "history"
-  const hasHeritageOrHistory = (array) => {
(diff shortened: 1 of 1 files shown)


src/gatsby/src/components/park/about.js
```diff
@@ -49,15 +49,21 @@ export const AccordionList = ({ eventKey, data, openAccordions, toggleAccordion,
           <HtmlContent>
             {!isNullOrWhiteSpace(data.description) && data.description}
           </HtmlContent>
-          {audioClip.length > 0 && (data.code === "heritage" || data.code === "history") ? (
+          {heritageAudioClip.length > 0 && data.code === "heritage" && (
+            <AudioButton
+              audio={heritageAudioClip[0]}
```
duncan-oxd, line 54: If we're always just using the first item from these arrays, we could use `.find` instead of `.filter` to just get the first match, and then we wouldn't have to keep filtering through the whole array. And we wouldn't need to check the length and use `[0]`

duncan-oxd (approved):
Looks good! I put in an idea for a possible small optimization if it makes sense.

I guess we've got a lot of arrays and a lot of filtering and finding with Strapi :D

PR #1630, opened 2025-05-30
CMS-823: Remove isFacilityOpen from filtering

ayumi-oxd:
### Jira Ticket:
CMS-123

### Description:
- Partial revert of https://github.com/bcgov/bcparks.ca/pull/1607
- Remove `isFacilityOpen`, `isActivityOpen`, and `isCampingOpen` from filtering

### DIFF
--- a/src/gatsby/src/utils/subAreaHelper.js
+++ b/src/gatsby/src/utils/subAreaHelper.js
@@ -47,9 +47,9 @@ const combineCampingTypes = (campings, campingTypes, subAreas) => {
   let arr = [];
   let obj = subAreas;
 
-  // filter the campings to include only active and open campings
+  // filter the campings to include only active
   const parkCampingTypes = campings.filter(
-    (camping) => camping.isActive && camping.isCampingOpen
+    (camping) => camping.isActive
   )
   // add the parkCampingTypes to the common object
   for (const parkCampingType of parkCampingTypes) {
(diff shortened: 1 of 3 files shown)


src/gatsby/src/utils/subAreaHelper.js
```diff
@@ -80,9 +80,9 @@ const combineFacilities = (facilities, facilityTypes, subAreas) => {
   let arr = [];
   let obj = subAreas;
 
-  // filter the facilities to include only active and open facilities
+  // filter the facilities to include only active
   const parkFacilities = facilities.filter(
-    (facility) => facility.isActive && facility.isFacilityOpen
```
duncan-oxd, line 85: The filter used to look at `isActive` _and_ `isFacilityOpen` - so they'd both have to be true to show on the page, right? It sounded like the bug ticket was saying some facilities were showing when they shouldn't be:
> Facilities "isActive" FALSE showing 

Is there any code that would allow that to happen? These changes look good, but if something was showing before, it would still be showing now, unless I'm missing something 👀

ayumi-oxd: If you look at https://github.com/bcgov/bcparks.ca/pull/1607, there was no filtering before. So all the facilities were displayed. That's why I added filtering in the previous PR, which were `isActive` and `isFacilityOpen`. However, `isFacilityOpen` should not control whether facilities are displayed.

duncan-oxd: Ah, just the title of the CMS-823 ticket was out of date 👍 Thanks for clarifying!

duncan-oxd (approved):

PR #1638, opened 2025-08-08
CMS-1159: Fix api error with park access status

molund:
### Jira Ticket:
CMS-1159

### Description:
This fixes an API error with /api/park-access-statuses

The code was trying to sort by precendence, but precedence wasn't included in the query.

It appears that this never worked properly on the API, although park-access-status was working on the website. 

In order to replicate the error you need to have a park with two advisories with access-status.precedence lower than 120, and the less important advisory needs to be older than the higher importance advisory.


ayumi-oxd (approved):
Good catch! so `precedence` wasn't really populated ... I didn't realize this until today 😅

PR #1639, opened 2025-08-26
CMS-1150: Swap homepage hero png for optimized jpg

duncan-oxd:
### Jira Ticket:
CMS-1150

### Description:

A small change to optimize the 6.6MB mountain photo on the homepage, since it's very big and one of the first things on the page. 

The project uses **gatsby-plugin-image** in a couple of places, and I think we should use it more, but for this quick fix I just optimized the file itself. It sounds a bit awkward to use the plugin for a CSS background image:
https://www.gatsbyjs.com/docs/how-to/images-and-media/using-gatsby-plugin-image/#background-images Let me know if you have other ideas/suggestions!

I was going to use an `image-set` with a jpg and AVIF and Webp for maximum compatibility but in the end, the Jpg was the smallest file, so I just used that.

I also noticed a bunch of other unoptimized PNGs on the homepage in the LinkCard components, but they're not as big so I'll put a ticket in the backlog to optimize those, possibly with the gatsby plugin.


duncan-oxd:
The new image is ~0.3MB instead of 6.6MB 👍

ayumi-oxd (approved):

PR #1640, opened 2025-09-05
CMS-836: Strapi: Create park-area and park-area-type collection

molund:
### Jira Ticket:
CMS-836

### Description:
Created new collections as an initial step in data restructuring.

### DIFF
--- a/src/cms/src/api/park-area/content-types/park-area/schema.json
+++ b/src/cms/src/api/park-area/content-types/park-area/schema.json
@@ -1,0 +1,213 @@
+{
+  "kind": "collectionType",
+  "collectionName": "park_areas",
+  "info": {
+    "singularName": "park-area",
+    "pluralName": "park-areas",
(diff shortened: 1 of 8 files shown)


ayumi-oxd (changes requested):
Sorry for leaving comments again, I noticed one thing to change 🙏

src/cms/src/api/park-area/content-types/park-area/schema.json
```diff
@@ -0,0 +1,213 @@
...
+    },
+    "yurt": {
+      "type": "string"
+    },
+    "shelters": {
+      "type": "string"
+    },
+    "boatLaunches": {
+      "type": "string"
+    },
```
ayumi-oxd, line 167: Should we change these fields to be "integer"? Copilot caught it on my PR: https://github.com/bcgov/bcparks.ca/pull/1644. It'd make sense since we store the number in these fields.

PR #1771, opened 2026-03-27
CMS-1605: Update image positioning/caption styles

duncan-oxd:
include images narrower than the content area

### Jira Ticket:
CMS-1605

### Description:

This branch fixes the CSS for positioning images in Strapi and on the frontend to handle a scenario found during UAT in CMS-1605. If an image is smaller than the content area (like a logo graphic), and it doesn't need to be resized, then it only has the `.image` class but not `.image_resized` so some of the styles wouldn't affect it. This branch changes the main selector to target `.image`.

I also updated the "figure" styles to use `display: table` and `table-caption` for `figcaption` because that's what the built-in CKEditor styles were doing. That's kind of a compatibility hack with older browsers, and we could do it with `fit-content` but I figured it's best to stay consistent with what CKEditor is doing here, since they designed the class/element structure.

### DIFF
--- a/src/gatsby/src/styles/_content-mixins.scss
+++ b/src/gatsby/src/styles/_content-mixins.scss
@@ -141,11 +141,8 @@
 
   figure {
     margin: 0 0 1em;
-
-    // Position table captions below the table to match the image captions
-    &.table figcaption {
-      caption-side: bottom;
-    }
+    // Use table display to match what CKEditor does with figures and captions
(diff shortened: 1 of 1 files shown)


duncan-oxd:
Tagging this one with 3.1.0 for inclusion in the next round of work after the Strapi 5 upgrade

ayumi-oxd (approved):

PR #1775, opened 2026-04-10
CMS-1654: Upgrade CKEditor Strapi plugin

duncan-oxd:
### Jira Ticket:
CMS-1654

### Description:

Upgraded the Strapi CKEditor plugin from 7.1.0 to 7.1.1, which should fix some visual bugs noticed by QA in the past, especially CMS-1544. 
I also notice this version bumps its CKEditor dependency from ~47.4.0 to ~47.6.1 so we might benefit from that upgrade as well. 
It's such a small ticket, I'd like to bundle it with #1771 when I send it to QA. It would fix a bunch of "known shippable" bugs from our Strapi upgrade effort, and they can go in a hotfix any time down the road.

Also removed an errant newline that was bugging me 😏 I checked the CKEditor stylesheet to make sure there were no workarounds for visual bugs in there.


molund:
Are you missing a package.json file from your commit?

duncan-oxd:
`npm update` just updates the lockfile and leaves package.json alone these days. It's harmless to update package.json as well (by re-installing the package) but nobody running `npm install` would ever get that old version again thanks to package-lock

ayumi-oxd (commented):
Is it for v3.0.0?

duncan-oxd:
This will either be for a 3.0.1, or rolled into 3.1.0 eventually

ayumi-oxd (approved):
I guess it'll be merged into alpha?
It might be worth updating package.json as well for clarity if the library update was intentional 🙂

duncan-oxd:
Ooh, good catch - I updated it to merge into alpha instead of main 👍 

The package.json file specifies the allowed range ("^7.1.0" means "latest 7.x.x, >= 7.1.0"), so it doesn't need to change for this upgrade. The exact installed version is pinned in package-lock.json along with all sub-dependencies.

If we rely on package.json to reflect the exact installed version, I'd be concerned that it won’t always be accurate unless we update it every time. We can use `npm ls <package-name>` to check the actual versions.

I'm happy to update the range in package.json if we want it to look more intentional, but it’s not required. Let me know what the team prefers!

PR #1777, opened 2026-04-15
CMS-1533: Update hasPageHeader condition for static content page

ayumi-oxd:
### Jira Ticket:
CMS-1533

### Description:
- Updated `hasPageHeader` condition for the static content page
  - Unlike the static landing page, the page header doesn't need both title and intro text
- Update Strapi version to v5.42.1

### DIFF
--- a/src/gatsby/src/templates/staticContent1.js
+++ b/src/gatsby/src/templates/staticContent1.js
@@ -58,7 +58,7 @@ export default function StaticContent1({ pageContext }) {
   // New non-repeatable page header component
   const pageHeader = page?.PageHeader || null
   const hasPageHeader =
-    pageHeader?.pageTitle &&
+    pageHeader?.pageTitle ||
     pageHeader?.introHtml?.data?.introHtml.length > 0
   const hasPageHeaderIntro = (
     pageHeader?.introHtml?.data?.introHtml.length > 0 || headerContent?.introHtml?.data?.introHtml.length > 0)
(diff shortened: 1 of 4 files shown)


src/gatsby/src/templates/staticContent1.js
```diff
@@ -58,7 +58,7 @@ export default function StaticContent1({ pageContext }) {
   // New non-repeatable page header component
   const pageHeader = page?.PageHeader || null
   const hasPageHeader =
-    pageHeader?.pageTitle &&
+    pageHeader?.pageTitle ||
     pageHeader?.introHtml?.data?.introHtml.length > 0
```
duncan-oxd, line 62: This change will work, but I noticed `hasPageHeader` is used throughout the template with a bunch of checks like this:

```jsx
{(hasPageHeader && hasPageHeader !== null) ? (...)}
```

The null check is redundant because if the value _is_ null, it won't get past the first truthy check. 

It looks like `hasPageHeader` is meant to be a boolean (name is `has...`) but it can currently be any value. Let's change the definition to be a boolean. Then we can simplify all of the checks that read its value.

Simplest solution is to just wrap this whole expression in `Boolean(...)`

duncan-oxd (changes requested):
I had a thought about clarifying how `hasPageHeader` is used and suggested a small tweak, plus some optional cleanup 👍

duncan-oxd (approved):
Nice 👍 the `hasPageHeader` variable is easier to figure out at a glance now

PR #1781, opened 2026-04-27
CMS-1696: Clear park-operation cache

ayumi-oxd:
### Jira Ticket:
CMS-1696

### Description:
- I'm not entirely sure about this approach, so any feedback is welcome
- If any fields in `park-operation` are updated, the `protected-area` API should be able to return the latest `park-operation` data accordingly

### DIFF
--- a/src/cms/src/middlewares/rest-cache-invalidation.js
+++ b/src/cms/src/middlewares/rest-cache-invalidation.js
@@ -11,6 +11,11 @@ const cachedCollectionTypes = [
   "api::park-feature.park-feature",
 ];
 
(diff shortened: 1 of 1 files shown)


molund (approved):
This should fix the issue, but there are several other collectionTypes we could also watch.

Anything with a relation to protected-area could qualify, but these would probably be the most likely to change frequently

- api::park-area.park-area
- api::park-gate.park-gate
- api::park-date.park-date
- api::park-facility.park-facility
- api::park-activity.park-activity
- api::park-camping-type.park-camping-type

PR #1783, opened 2026-04-28
CMS-1672: Seed rank columns for feature types and park area types

molund:
### Jira Ticket:
CMS-1672
CMS-1673

### Description:
This PR adds a migration to Strapi that imports the data in the spreadsheet attached to CMS-1673 into Strapi
https://bcparksdigital.atlassian.net/browse/CMS-1673

- The `sortKey` column from the `AreaType-FeatureTypeGrouping` tab is being used as the rank. 
- The spreadsheet was incomplete, so some additional ParkAreas had to be added by following the pattern in the spreadsheet.

This is set to be merged into the main branch because it’s better to have the data on production where it can be centrally managed, rather than leaving it on the alpha environment while the imported spreadsheet data becomes stale.

### DIFF
--- a/src/cms/database/migrations/2026.04.28T00.00.007-set-feature-and-area-type-ranks.js
+++ b/src/cms/database/migrations/2026.04.28T00.00.007-set-feature-and-area-type-ranks.js
@@ -1,0 +1,161 @@
+"use strict";
+
+module.exports = {
+  async up(knex) {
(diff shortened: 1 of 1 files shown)


ayumi-oxd (approved):
Looks good! Just left a question about the target branch: https://github.com/bcgov/bcparks-staff-portal/pull/535#issuecomment-4347406536

PR #1812, opened 2026-05-11
CMS-1762: Update slug character validator regex

ayumi-oxd:
### Jira Ticket:
CMS-1762

### Description:
- It failed because `slugCharacterValidator` requires the first character to be `[a-z0-9]` but the slugs in the page collection have a leading `/`
- Updated it to accept the first char may be `/` and `[a-z0-9]`

### DIFF
--- a/src/cms/src/helpers/validator.js
+++ b/src/cms/src/helpers/validator.js
@@ -42,7 +42,7 @@ module.exports = {
   },
   // checks for valid characters and consecutive forward slashes
   slugCharacterValidator: function (slug) {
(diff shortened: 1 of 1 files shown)


molund (approved):
Looks good to me. Were you able to determine why this broke all of a sudden?

ayumi-oxd:
> Looks good to me. Were you able to determine why this broke all of a sudden?

@molund I guess we just didn’t notice this bug until recently. The only slug that requires a “/” is this page slug, and we rarely update those page entries.

PR #1814, opened 2026-05-13
NOBUG: Fix gatsby build ssr bundle error

ayumi-oxd:
### Jira Ticket:
None

### Description:
- There is a Gatsby build error on the alpha environment. Note that it does not happen on the main.
```
error Generating SSR bundle failed

Reading from "node:sqlite" is not handled by plugins (Unhandled scheme).
Webpack supports "data:" and "file:" URIs by default.
You may need an additional plugin to handle "node:" URIs.
not finished Running gatsby-plugin-sharp.IMAGE_PROCESSING jobs - 30.314s
not finished Building HTML renderer - 7.615s
```
  - It looks like one dependency chain cheerio -> undici uses imported `node:sqlite`
  - Webpack in Gatsby SSR did not handle that `node:` scheme by default so the error happened

### DIFF
--- a/src/gatsby/gatsby-node.js
+++ b/src/gatsby/gatsby-node.js
@@ -2,7 +2,6 @@ const { graphql } = require("gatsby");
 const slugify = require("slugify");
 const NodePolyfillPlugin = require("node-polyfill-webpack-plugin");
 const parseUrl = require("parse-url");
-const webpack = require("webpack");
 
 const strapiApiRequest = (graphql, query) =>
   new Promise((resolve, reject) => {
@@ -712,25 +711,10 @@ async function createRedirects(
 exports.onCreateWebpackConfig = ({ stage, actions, getConfig }) => {
   if (stage === "build-html" || stage === "develop-html") {
     actions.setWebpackConfig({
-      resolve: {
-        alias: {
-          "node:sqlite": false,
-        },
-        fallback: {
-          sqlite: false,
-        },
-      },
       module: {
         rules: [],
       },
-      plugins: [
-        new webpack.NormalModuleReplacementPlugin(/^node:/, (resource) => {
-          resource.request = resource.request.replace(/^node:/, "");
-        }),
-        new webpack.IgnorePlugin({ resourceRegExp: /^sqlite$/ }),
-        new webpack.IgnorePlugin({ resourceRegExp: /^node:sqlite$/ }),
-        new NodePolyfillPlugin(),
-      ],
+      plugins: [new NodePolyfillPlugin()],
     });
   }
   if (stage === "build-javascript" || stage === "develop") {
--- a/src/gatsby/gatsby-node.js
+++ b/src/gatsby/gatsby-node.js
@@ -2,6 +2,7 @@ const { graphql } = require("gatsby");
 const slugify = require("slugify");
 const NodePolyfillPlugin = require("node-polyfill-webpack-plugin");
 const parseUrl = require("parse-url");
+const webpack = require("webpack");
 
 const strapiApiRequest = (graphql, query) =>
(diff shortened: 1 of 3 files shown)


molund:
It looks like it was prettier that imported the dependency.

src/gatsby/gatsby-node.js
```diff
@@ -711,10 +712,25 @@ async function createRedirects(
...
+          sqlite: false,
+        },
+      },
       module: {
         rules: [],
       },
-      plugins: [new NodePolyfillPlugin()],
+      plugins: [
+        new webpack.NormalModuleReplacementPlugin(/^node:/, (resource) => {
+          resource.request = resource.request.replace(/^node:/, "");
```
molund: Do we actually need to rewrite all node: imports, or just block sqlite?”

duncan-oxd:
That's weird, the "node:" prefix on a module name is just used to indicate that it's a core module (like `node:fs`) but that's existed since long before v22 (which the builder uses) 🤔 
Is there an update for webpack (or one of its loaders) that we can apply to fix this?

duncan-oxd:
This might be related, using the "externals" plugin: https://github.com/webpack/webpack/issues/14798#issuecomment-976225678
(Though this is quite an old post)

ayumi-oxd:
> It looks like it was prettier that imported the dependency.

@molund In this change? https://github.com/bcgov/bcparks.ca/commit/bee10fd269fbfbe7081b213580242758de0c418e Or did you mean Copilot? 
It looks like that dependency chain cheerio -> undici uses `node:sqlite`.
Should I downgrade it to the same version as the main? Since it doesn't happen on the main 🤔

src/gatsby/gatsby-node.js
ayumi-oxd: Just block sqlite

ayumi-oxd:
> That's weird, the "node:" prefix on a module name is just used to indicate that it's a core module (like `node:fs`) but that's existed since long before v22 (which the builder uses) 🤔 Is there an update for webpack (or one of its loaders) that we can apply to fix this?

@duncan-oxd It looks like it's not a webpack thing, it seems that dependency chain cheerio -> undici uses node:sqlite. But I'm not entirely sure 😅

duncan-oxd:
Using that core module should be no problem, but it sounds like Webpack doesn't like the syntax for some reason. (You can import and use "node:sqlite" in your own code any time). I think Webpack sees the "node:" part and thinks it needs to use a special loader, but it should be just using the built-in sqlite module 🤔 After a google search, it looks like many people have experienced this but I didn't see a definite solution at a quick glance... Changing the config for the "externals" plugin sounds reasonable though.

Note that the `node:sqlite` module was added in Node 22 and it was marked experimental at the time, so maybe our version of Webpack has some internal whitelist of core modules that doesn't include sqlite yet, so it thinks it's an external package..?

molund:
> > It looks like it was prettier that imported the dependency.
> 
> @molund In this change? [bee10fd](https://github.com/bcgov/bcparks.ca/commit/bee10fd269fbfbe7081b213580242758de0c418e) Or did you mean Copilot? It looks like that dependency chain cheerio -> undici uses `node:sqlite`. Should I downgrade it to the same version as the main? Since it doesn't happen on the main 🤔

@ayumi-oxd this one seems more likely: https://github.com/bcgov/bcparks.ca/commit/7bc18350b062070e783e53b5eda6973a98a32ec6

It's getting the prettier info from the cms project and using it globally.  I did it this way because I was trying to fix an issue where different people's prettier version was resulting in different formatting rules, but it seems to have found it's way into node_modules for the gatby project too.

molund (approved):

PR #1815, opened 2026-05-13
CMS-1737: Add sort order to advisory status

ayumi-oxd:
### Jira Ticket:
CMS-1737

### Description:
- The new sorting for advisory status in the advisory table required a defined sort order
  - Added `precedence` field to advisory status
  - Added a migration script to populate `precedence` in order of order Draft > HQ review > Scheduled > Published > Unpublished

### DIFF
--- a/src/cms/database/migrations/2026.05.13T00.00.010-populate-advisory-status-sort-order.js
+++ b/src/cms/database/migrations/2026.05.13T00.00.010-populate-advisory-status-sort-order.js
@@ -6,21 +6,21 @@ module.exports = {
       return;
     }
 
-    if (!(await knex.schema.hasColumn("advisory_statuses", "precedence"))) {
+    if (!(await knex.schema.hasColumn("advisory_statuses", "sortOrder"))) {
       await knex.schema.table("advisory_statuses", (table) => {
-        table.integer("precedence");
+        table.integer("sortOrder");
       });
     }
 
     await knex.raw(`
       UPDATE advisory_statuses
-      SET precedence = CASE advisory_status
+      SET sortOrder = CASE advisory_status
         WHEN 'Draft' THEN 0
         WHEN 'HQ review' THEN 1
         WHEN 'Scheduled' THEN 2
         WHEN 'Published' THEN 3
         WHEN 'Unpublished' THEN 4
-        ELSE precedence
+        ELSE sortOrder
       END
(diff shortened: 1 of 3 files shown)


src/cms/src/api/advisory-status/content-types/advisory-status/schema.json
```diff
@@ -18,6 +18,11 @@
     },
     "code": {
       "type": "string"
+    },
+    "precedence": {
+      "type": "integer",
+      "unique": true,
+      "required": false
```
molund: It should be safe to make this required. your migration is creating the column and populating it before the schema synchronization occurs.
Strapi actually doesn't enforce NOT NULL or UNIQUE constraints at the db level so you can get a away with quite a bit.

duncan-oxd (approved):
Looks good to me! I think some of the other "sort order" fields use larger numbers like 10, 20, 30 so we can easily add new items in between, but with only 5 items it's probably easy enough to update a few rows at once (if we ever need to)

src/cms/database/migrations/2026.05.13T00.00.010-populate-advisory-status-sort-order.js
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    if (!(await knex.schema.hasColumn("advisory_statuses", "precedence"))) {
+      await knex.schema.table("advisory_statuses", (table) => {
+        table.integer("precedence");
+      });
+    }
+
+    await knex.raw(`
+      UPDATE advisory_statuses
```
molund, line 16: precedence **isn't** a great name for this field because it implies priority, not just display order.  

We use `rank` for sort order on activityType, campingType,  facilityType, featureType and parkAreaType.

But I noticed standardMessage does use precedence as well.

src/cms/database/migrations/2026.05.13T00.00.010-populate-advisory-status-sort-order.js
```diff
@@ -6,21 +6,21 @@ module.exports = {
       return;
     }
 
-    if (!(await knex.schema.hasColumn("advisory_statuses", "precedence"))) {
+    if (!(await knex.schema.hasColumn("advisory_statuses", "sortOrder"))) {
       await knex.schema.table("advisory_statuses", (table) => {
-        table.integer("precedence");
+        table.integer("sortOrder");
```
molund: Strapi will rename sortOrder to sort_order when it stores the field in the db. Knex code has to use actual Postgres column names.

PR #1817, opened 2026-05-14
CMS-1731: Add unpublished fields to public advisory audit

ayumi-oxd:
### Jira Ticket:
CMS-1731

### Description:
- In the Advisory Review tab, advisories that have been moved to the unpublished state will be displayed in the following cases:
  - They are automatically unpublished when they expire based on the expiry date.
  - They are manually unpublished when someone clicks the “Unpublish” action.
- For the second case, the system currently does not track who unpublished an advisory or when it was intentionally unpublished. To address this, I added these fields to the public advisory audit table to track manual unpublish actions.


duncan-oxd (approved):
Looks good, it follows the same naming conventions as the "createdByName" / "createdAt" fields (and I did the same for the "reviewed..." fields 👍

PR #1824, opened 2026-05-15
NOBUG: Prevent Webpack transforming core modules, update cheerio

duncan-oxd:
### Jira Ticket:
(No Jira ticket)

### Description:
Optional follow-up for #1814 - this upgrades the Cheerio dependency again and uses the Webpack externals plugin to tell webpack not to try to transform/polyfill node core modules like `node:sqlite` (and also the `undici`, the package that caused the issue).

Webpack sees a package importing `node:sqlite` and thinks it needs to do something with the `node:` URL scheme, when it's supposed to recognize it as a built-in Node package. 

> Reading from "node:sqlite" is not handled by plugins (Unhandled scheme).
Webpack supports "data:" and "file:" URIs by default.
You may need an additional plugin to handle "node:" URIs.

Webpack 5.98.0 is new enough to know about `node:sqlite` (added in Node v22.5), but it looks like it's not doing that when it builds the Gatsby HTML renderer bundle. 

This webpack config change tells it to explicitly treat it as an external module and not bundle or transform it, so Node just handles it at runtime.

The build succeeds locally with this setup, but we should explicitly test whatever parts of the app are using cheerio (the Strapi link decorators, the pre-render-video hook, Park Overview pg., etc.)

### DIFF
--- a/src/gatsby/gatsby-node.js
+++ b/src/gatsby/gatsby-node.js
@@ -708,15 +708,31 @@ async function createRedirects(
   });
 }
 
+// Customize the webpack config for specific build stages
 exports.onCreateWebpackConfig = ({ stage, actions, getConfig }) => {
+  // Server-side HTML renderer bundle
   if (stage === "build-html" || stage === "develop-html") {
     actions.setWebpackConfig({
       module: {
         rules: [],
       },
       plugins: [new NodePolyfillPlugin()],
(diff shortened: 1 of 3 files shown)


ayumi-oxd (approved):
Looks good!

molund (approved):

PR #1843, opened 2026-05-28
CMS-1733: Add createdByEmail and contibutor role to public-advisory-audits

molund:
### Jira Ticket:
CMS-1733

### Description:

- Added new createdByEmail field to the Strapi public-advisory-audits collection
- Expanded email notifiatitions for after hours advisories to include the "contributor" role (waiting for clarifiation from UX about expected behaviour)


ayumi-oxd (approved):
Looks good, but what is the difference between "submitter" and "contributor"?

molund:
> Looks good, but what is the difference between "submitter" and "contributor"?

Right now there is no group that is only a contributor, but according to CMS-1640 a contributor should only be able to save as draft.

PR #1851, opened 2026-06-04
CMS-1792: Change api page size max limit

ayumi-oxd:
### Jira Ticket:
CMS-1792

### Description:
- Increased the Strapi API maximum page size limit from 2,000 to 4,000
- When a user selected "All" rows per page in the frontend, only 2,000 advisories were returned, causing some advisories to be hidden unless the user performed a search
- There are currently 3,070 records in `public-advisory-audit` with `isLatestRevision="true"`, so the maximum limit has been increased to 4,000 to accommodate all current records

### DIFF
--- a/src/cms/config/api.js
+++ b/src/cms/config/api.js
@@ -4,6 +4,6 @@
 module.exports = {
   rest: {
     defaultLimit: 100,
-    maxLimit: 2000,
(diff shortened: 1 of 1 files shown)


ayumi-oxd:
Closing this PR, as we've decided not to display all advisories at once.

PR #1857, opened 2026-06-09
CMS-1751: Set access-status scope

molund:
### Jira Ticket:
CMS-1751

### Description:
This change updates the access status scope values to match the spreadsheet in the Jira ticket.

### DIFF
--- a/src/cms/database/migrations/2026.06.08T00.00.016-access-statuses.js
+++ b/src/cms/database/migrations/2026.06.08T00.00.016-access-statuses.js
@@ -1,0 +1,30 @@
+"use strict";
+
+/*
+This migration updates the scope field on existing records in the access_statuses
+table to be "BCP" or "Both" based data provided in a spreadsheet in Jira.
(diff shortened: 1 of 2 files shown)


ayumi-oxd (commented):
Access status doesn't need a category field?

molund:
> Access status doesn't need a category field?

`groupLabel` is the category field

ayumi-oxd:
> > Access status doesn't need a category field?
> 
> `groupLabel` is the category field

Understood, does it need to be an enum field, the same as `eventType`?

molund:
> > > Access status doesn't need a category field?
> > 
> > 
> > `groupLabel` is the category field
> 
> Understood, does it need to be an enum field, the same as `eventType`?

That's a good idea. I added an emum and I also deleted the inactive record that has never been used because I didn't want to pollute the enum with an extra value.

ayumi-oxd (approved):
Looks good!

PR #1859, opened 2026-06-09
CMS-1795: Update email triggers and subjects

duncan-oxd:
### Jira Ticket:
CMS-1795

### Description:

Changes to the email scheduler service, in preparation for updating some emails and adding a couple of new ones in future tickets.

- Selectively add the advisory headline to the email subject instead of automatically doing it every time (some upcoming emails won't include the headline)
- Updated several email subjects to include mention of closures
- Added "additional recipients" support to add more "to" email addresses for some messages. This is used to include the original poster on several of the emails.

Updated email trigger conditions:
"Expiring soon" - sends out 1 week before expiry date (previously was 2 and 5 days before)

# For discussion
The task also mentions a few things we might want to change or implement as part of this ticket (or its related tickets)

> Need to ensure the dev and test environments do not spam Parks staff with emails.
> Perhaps via whitelists.
> Do not include emails in public API (should only be in public-advisory-audit).

### DIFF
--- a/src/cms/src/api/public-advisory/services/scheduling.js
+++ b/src/cms/src/api/public-advisory/services/scheduling.js
@@ -6,82 +6,196 @@
 
 const { queueAdvisoryEmail } = require("../../../helpers/taskQueue.js");
 
-module.exports = ({ strapi }) => ({
-  expire: async (advisoryStatusMap) => {
-    let expiredAdvisoryCount = 0;
+/**
+ * Adds the advisory headline to the email subject.
+ * @param {string} subject the base subject for the email
+ * @param {object} advisoryData the advisory data
+ * @param {string} advisoryData.title the title of the advisory
+ * @return {string} subject with headline appended if it exists
+ */
+function addHeadlineToSubject(subject, advisoryData) {
+  if (advisoryData.title) {
+    return `${subject}: ${advisoryData.title}`;
+  }
 
-    if (Object.keys(advisoryStatusMap).length > 0) {
-      // fetch advisories to unpublish - public advisory table
-      const advisoryToUnpublish = await strapi
-        .documents("api::public-advisory.public-advisory")
-        .findMany({
-          filters: {
-            expiryDate: {
-              $lte: new Date().toISOString(),
-            },
-            advisoryStatus: advisoryStatusMap["PUB"].id,
-          },
-          populate: "*",
-        });
+  // Return the original subject if no headline exists
+  return subject;
+}
 
-      // delete advisories - public advisory table
-      for (const advisory of advisoryToUnpublish) {
-        strapi.log.info(
-          `unpublishing public-advisory [advisoryNumber:${advisory.advisoryNumber}]`,
-        );
-        await strapi
+module.exports = ({ strapi }) => {
+  /**
+   * Gets the original creator email for an advisory.
+   * Returns the createdByEmail value from the first corresponding revision in public-advisory-audit.
+   * @param {number} advisoryNumber the advisory number to look up
+   * @returns {Promise<string|undefined>} creator email, if found
+   */
+  async function getCreatorEmail(advisoryNumber) {
+    // Find the original advisory revision to get the submitter email.
+    const advisoryAudit = await strapi
+      .documents("api::public-advisory-audit.public-advisory-audit")
+      .findFirst({
+        filters: {
+          advisoryNumber,
+          revisionNumber: 1,
+        },
+        fields: ["createdByEmail"],
+      });
+
+    return advisoryAudit?.createdByEmail;
+  }
+
+  return {
+    expire: async (advisoryStatusMap) => {
+      let expiredAdvisoryCount = 0;
+
+      if (Object.keys(advisoryStatusMap).length > 0) {
+        // fetch advisories to unpublish - public advisory table
+        const advisoryToUnpublish = await strapi
           .documents("api::public-advisory.public-advisory")
-          .update({
-            documentId: advisory.documentId,
-            data: {
-              publishedAt: null,
+          .findMany({
+            filters: {
+              expiryDate: {
+                $lte: new Date().toISOString(),
+              },
+              advisoryStatus: advisoryStatusMap["PUB"].id,
             },
-          })
-          .then(async (advisory) => {
-            expiredAdvisoryCount++;
-            await queueAdvisoryEmail(
-              "Expired advisory removed",
-              "An expired advisory was removed",
-              advisory.advisoryNumber,
-              "public-advisory-audit::services::scheduling::expire()",
-            );
-          })
-          .catch((error) => {
-            strapi.log.error(
-              `error updating public-advisory #${advisory.advisoryNumber}`,
-              error,
-            );
+            populate: "*",
           });
-      }
 
-      // unpublish advisories - audit table
-      for (const advisory of advisoryToUnpublish) {
-        const advisoryAudit = await strapi
+        // delete advisories - public advisory table
+        for (const advisory of advisoryToUnpublish) {
+          strapi.log.info(
+            `unpublishing public-advisory [advisoryNumber:${advisory.advisoryNumber}]`,
+          );
+          await strapi
+            .documents("api::public-advisory.public-advisory")
+            .update({
+              documentId: advisory.documentId,
+              data: {
+                publishedAt: null,
+              },
+            })
+            .then(async (advisory) => {
+              expiredAdvisoryCount++;
+
+              const subject = addHeadlineToSubject(
+                "Expired advisory / closure was removed",
+                advisory,
+              );
+
+              // Find the creator email from the original advisory-audit record
+              const creatorEmail = await getCreatorEmail(
+                advisory.advisoryNumber,
+              );
+
+              await queueAdvisoryEmail(
+                subject,
+                "An expired advisory was removed",
+                advisory.advisoryNumber,
+                "public-advisory-audit::services::scheduling::expire()",
+                creatorEmail ? [creatorEmail] : [],
+              );
+            })
+            .catch((error) => {
+              strapi.log.error(
+                `error updating public-advisory #${advisory.advisoryNumber}`,
+                error,
+              );
+            });
+        }
+
+        // unpublish advisories - audit table
+        for (const advisory of advisoryToUnpublish) {
+          const advisoryAudit = await strapi
+            .documents("api::public-advisory-audit.public-advisory-audit")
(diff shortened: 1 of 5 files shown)


src/scheduler/email-alerts/scripts/mailer.js
```diff
@@ -1,6 +1,17 @@
 const nodemailer = require("nodemailer");
 
-exports.send = async function (subject, body, summary, fromName) {
+// Combines the primary email recipient with any additional recipients.
+const buildRecipientList = function (additionalRecipients) {
+  return [process.env.EMAIL_RECIPIENT, ...additionalRecipients].filter(Boolean);
+};
```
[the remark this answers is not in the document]
duncan-oxd: I set the default param to [] for additionalRecipients in `send`

src/cms/src/api/public-advisory/services/scheduling.js
```diff
@@ -6,18 +6,37 @@
...
+}
+
 module.exports = ({ strapi }) => ({
   expire: async (advisoryStatusMap) => {
     let expiredAdvisoryCount = 0;
 
     if (Object.keys(advisoryStatusMap).length > 0) {
+      // Look ahead one minute to handle expiries on a minute-by-minute cron schedule
+      const expiryCutoff = new Date(Date.now() + 60 * 1000).toISOString();
+
```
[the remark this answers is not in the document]
duncan-oxd: I'll revert it to how it was.

src/cms/src/api/public-advisory/services/scheduling.js
```diff
@@ -39,11 +58,18 @@ module.exports = ({ strapi }) => ({
...
+            );
+
             await queueAdvisoryEmail(
-              "Expired advisory removed",
+              subject,
               "An expired advisory was removed",
               advisory.advisoryNumber,
               "public-advisory-audit::services::scheduling::expire()",
+              advisory.createdByEmail ? [advisory.createdByEmail] : [],
             );
```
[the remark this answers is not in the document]
duncan-oxd: I pushed an update to use a lookup map to find creator emails in these two cases

src/cms/src/api/public-advisory/services/scheduling.js
```diff
@@ -171,11 +204,18 @@ module.exports = ({ strapi }) => ({
...
+      );
+
       await queueAdvisoryEmail(
-        "Advisory expiring soon",
+        subject,
         "This advisory will be expiring in one week",
         advisory.advisoryNumber,
         "public-advisory::services::scheduling::expiringSoon()",
+        advisory.createdByEmail ? [advisory.createdByEmail] : [],
       );
```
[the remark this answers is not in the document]
duncan-oxd: I pushed an update to use a lookup map to find creator emails in these two cases

src/scheduler/email-alerts/scripts/sendAdvisoryEmails.js
```diff
@@ -123,7 +123,13 @@ exports.sendAdvisoryEmails = async function (recentAdvisoryEmails) {
             process.env.BCPARKS_ENVIRONMENT.toLowerCase() === "prod"
               ? "Staff Web Portal"
               : process.env.BCPARKS_ENVIRONMENT.toUpperCase();
-          await send(subject, htmlMessageBody, summary, fromName);
+          await send(
+            subject,
+            htmlMessageBody,
+            summary,
+            fromName,
+            emailInfo.additionalRecipients ?? [],
```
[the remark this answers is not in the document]
duncan-oxd: I think this is planning for a scenario that will never happen, but I'll commit the change anyway 🤷

src/scheduler/email-alerts/scripts/sendAdvisoryEmails.js
```diff
@@ -114,7 +114,7 @@ exports.sendAdvisoryEmails = async function (recentAdvisoryEmails) {
 
       if (scriptKeySpecified("emailsend") || noCommandLineArgs()) {
         if (process.env.EMAIL_ENABLED.toLowerCase() !== "false") {
-          const subject = `${emailData.subject}: ${emailData.data.title}`;
+          const subject = emailData.subject;
```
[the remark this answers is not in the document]
duncan-oxd: There are no other contexts that call sendAdvisoryEmails besides the scheduler files we've updated here, so this is fine.

src/cms/src/api/public-advisory/services/scheduling.js
```diff
@@ -6,6 +6,58 @@
...
+  // Find the latest audit record for each advisory number and get the creator email
+  const advisoryAudits = await strapi
+    .documents("api::public-advisory-audit.public-advisory-audit")
+    .findMany({
+      filters: {
+        advisoryNumber: { $in: advisoryNumbers },
+        isLatestRevision: true,
+      },
+      fields: ["advisoryNumber", "createdByEmail"],
+    });
```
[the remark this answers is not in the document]
duncan-oxd: We're passing isLatestRevision and a list of advisory numbers to the query, so it's unlikely to exceed 25 but I'll check with the team about what they recommend!

src/scheduler/email-alerts/scripts/sendAdvisoryEmails.js
```diff
@@ -123,7 +123,13 @@ exports.sendAdvisoryEmails = async function (recentAdvisoryEmails) {
             process.env.BCPARKS_ENVIRONMENT.toLowerCase() === "prod"
               ? "Staff Web Portal"
               : process.env.BCPARKS_ENVIRONMENT.toUpperCase();
-          await send(subject, htmlMessageBody, summary, fromName);
+          await send(
+            subject,
+            htmlMessageBody,
+            summary,
+            fromName,
+            emailInfo?.additionalRecipients ?? [],
```
molund: It would be cleaner to build the recipient list here to include `process.env.EMAIL_RECIPIENT` and then mailer.send() could be more generic.

src/cms/src/api/public-advisory/services/scheduling.js
```diff
@@ -6,6 +6,58 @@
...
+  return subject;
+}
+
+/**
+ * Builds a map of advisory numbers to creator email addresses from latest audit records.
+ * @param {Strapi} strapi Strapi instance
+ * @param {Array<{advisoryNumber?: number}>} advisories Advisory records to resolve
+ * @returns {Promise<Map<number, string>>} Map of advisory number to creator email
+ */
+async function getAuditCreatorEmailMap(strapi, advisories) {
```
molund: I think this might be unnecessary complexity.  We only ever send an email for one advisory at a time. On line 107 you could just get createdByEmail from the current advisory.

src/cms/src/api/public-advisory/services/scheduling.js
```diff
@@ -24,6 +76,12 @@ module.exports = ({ strapi }) => ({
           populate: "*",
         });
 
+      // Get creator email addresses for each advisory to unpublish
+      const creatorEmailMap = await getAuditCreatorEmailMap(
+        strapi,
```
molund: You don't need to pass in `strapi` to helper functions. It's available from outer scope.

src/cms/src/helpers/taskQueue.js
```diff
@@ -84,6 +84,7 @@ module.exports = {
     title,
```
molund, line 84: It would be nice if we could get UX to make the subjects the same as the title prefixes so we wouldn't have to maintain both. I don't remember them being different before.

src/cms/src/api/public-advisory/services/scheduling.js
duncan-oxd: I used the createdByEmail directly where it's available, but for a few of the emails (like expiring published advisories), it queries the public advisories collection instead of the audit collection, so the creator email isn't available. It's also doing a "findMany" in these cases, so I set it up as a lookup map but I'd definitely like to simplify this if possible. Any ideas?

src/cms/src/api/public-advisory/services/scheduling.js
duncan-oxd: Ah, I defined this function right outside of that scope, but if I move it inside, I can get rid of that strapi parameter. 👍 I just pushed an update that defines the function inside the scope where `strapi` is available

src/scheduler/email-alerts/scripts/sendAdvisoryEmails.js
duncan-oxd: I pushed an update that renames that param to "recipients" and now we always pass it in: an array with the usual EMAIL_RECIPIENT, and sometimes the additional recipients passed to `queueAdvisoryEmail`

src/cms/src/helpers/taskQueue.js
duncan-oxd: There are some future tickets in this sprint for updating the contents, I just included the subject lines in this one, so there's a chance it will align. We'll see if anything can be refactored when I get to the later tickets 👍

src/cms/src/api/public-advisory/services/scheduling.js
molund: I didn't notice those usages were using public-advisories and not public-advisory-audits.

You could do the lookup inside the loop and get the value for revisionNumber 1.  The loop is almost never going to have more than one iteration anyway. 

```
const advisoryAudit = await strapi
  .documents("api::public-advisory-audit.public-advisory-audit")
  .findFirst({
    filters: {
      advisoryNumber,
      revisionNumber: 1,
    },
    fields: ["createdByEmail"],
  });

const createdByEmail = advisoryAudit?.createdByEmail;
```

duncan-oxd: That's a nice simplification 👍 I pushed a commit to update that (and rebased onto the latest changes in alpha)
Testing it locally, it seems to work with an expiring RST closure:
```
{
  "title": "An expired advisory was removed",
  "subject": "Expired advisory / closure was removed: Expiring at 2:45pm",
  "triggeredBy": "public-advisory-audit::services::scheduling::expire()",
  "advisoryNumber": 3428,
  "additionalRecipients": [
    "<my bcgov email from my IDIR>@gov.bc.ca"
  ]
}
```

molund (approved):

duncan-oxd:
Following our chat this morning, I pushed a small update to split EMAIL_RECIPIENT on the comma in case it's a comma-separated list of multiple values, as it is on some of our test environments.

duncan-oxd:
Rebased onto alpha to include the changes from #1861 - I'll merge now 👍

PR #1861, opened 2026-06-11
NOBUG: Update unpublishedByName when advisory gets expired

ayumi-oxd:
### Jira Ticket:
None

### Description:
- Updated `unpublishedByName: "system"` and `unpublishedDate` when an advisory gets unpublished automatically because of `expiryDate`
- Removed `removalDate`


molund (approved):

molund:
I think you are right. `removalDate` doesn't appear to be used anymore.  I already removed it from `public-advisories` but I didn't check usage of the field in `public-advisory-audits` at the time. 

Maybe you can push another commit to this PR that removes it? It looks like there are no references in the other repo.

PR #1862, opened 2026-06-12
CMS-1827: Create public-advisory-audits/scheduled-and-published endpoint

molund:
### Jira Ticket:
CMS-1827

### Description:

Added a scheduledAndPublished action to the public-advisory-audit controller. It returns all published audit rows matched by advisoryNumber/revisionNumber pair against the live advisories table, plus any SCH rows where isLatestRevision is true. Caller-supplied querystring filters are merged with a logical AND, and standard Strapi query params (fields, populate, sort, pagination) are fully supported.

RecSpace and Team Orca need a reliable way to pull currently published and scheduled advisory revisions from the audit table. The pair-match approach is necessary because the latest audit revision may be newer than the published one.

Replicating this logic via the standard Strapi REST API would require multiple round trips — first to fetch the live published advisories, then to query the audit table per pair — along with client-side merging and deduplication. The custom endpoint handles all of this server-side in a single call.

e.g. http://localhost:1337/api/public-advisory-audits/scheduled-and-published?populate[0]=accessStatus&populate[1]=eventType&populate[2]=advisoryStatus&populate[3]=recreationResources&populate[4]=urgency&filters[recreationResources][id][$notNull]=true

Note: By default the new endpoint will be private. It will need to be made accessible to the authenticated role via a post deployment task.

### DIFF
--- a/src/cms/src/api/public-advisory-audit/controllers/public-advisory-audit.js
+++ b/src/cms/src/api/public-advisory-audit/controllers/public-advisory-audit.js
@@ -35,7 +35,7 @@ module.exports = createCoreController(
       const { filters, sort, populate, fields, pagination } = ctx.query;
 
       // Get all the published public advisories from the projection table.
-      // We need to do it this way because it's possible for the lastest revision
+      // We need to do it this way because it's possible for the latest revision
       // of a public advisory audit to be newer than the published revision.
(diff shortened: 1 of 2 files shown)


diego-oxd (approved):
Yeah, this looks great!

PR #1863, opened 2026-06-15
CMS-1798: Email template updates

duncan-oxd:
### Jira Ticket:
CMS-1798

### Description:

Updates to the ACT email contents from Figma frame #0.0, Item numbers 1-3 and letters A-B

Updated the template with a new logo graphic and header text.

The emails all use the same template but they now have different fields displaying at the bottom, so I've added a new "metadata" field to specify which fields to display there, and in what order. An alternative would be to just give each email some kind of identifier and use that to sort it out (or use that identifier to load a different template for each email), but the templates are all the same otherwise.


src/cms/src/middlewares/staff-portal-advisory-audit.js
```diff
@@ -50,12 +51,18 @@ module.exports = () => {
 
     const newAdvisoryStatus = newPublicAdvisoryAudit.advisoryStatus?.code;
 
+    const urgency =
+      newPublicAdvisoryAudit.urgency?.urgency?.toLowerCase() ?? "";
+    const subject = `Review draft: ${urgency} urgency advisory / closure`;
```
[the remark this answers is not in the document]
duncan-oxd: The UI has it so the Urgency field will never be null or undefined, so this isn't necessary

src/cms/src/middlewares/staff-portal-advisory-audit.js
```diff
@@ -192,12 +201,17 @@ module.exports = () => {
     const oldAdvisoryStatus = ctx.state?.oldStatus; // saved by beforeUpdate() above
     const newAdvisoryStatus = publicAdvisoryAudit.advisoryStatus?.code;
 
+    const urgency = publicAdvisoryAudit.urgency?.urgency?.toLowerCase() ?? "";
+    const subject = `Review draft: ${urgency} urgency advisory / closure`;
```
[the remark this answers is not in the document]
duncan-oxd: Same as above, urgency will always be defined so an awkward email subject would be the least of our problems if it was missing here

src/scheduler/email-alerts/scripts/mailer.js
```diff
@@ -14,6 +14,7 @@ exports.send = async function (
   summary,
   fromName,
   recipients,
+  attachments = [],
```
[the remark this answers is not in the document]
duncan-oxd: Will do 👍

src/scheduler/email-alerts/templates/public-advisory.ejs
```diff
@@ -7,16 +7,18 @@
 </head>
 
 <body style="font-family: Arial, Helvetica; color: #1a1a1a">
-  <img src="cid:logo.png" alt="BC Parks" title="BC Parks Logo" style="margin: 8px 16px 0px 0px;" width="150" />
+  <img src="cid:logo.png" alt="BC Parks" title="BC Parks Logo" style="margin: 0 8px 0 0" width="150" />
 
   <span style="display:inline-block;font-size:16pt;white-space:nowrap;margin-top:10px;color:#003366;">
-    &nbsp;Staff web portal
+    &nbsp;Parks and Recreation staff web portal
```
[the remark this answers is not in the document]
duncan-oxd: Oops, done

src/scheduler/.prettierignore
duncan-oxd, line 1: To prevent Prettier from mangling these EJS templates on save

ayumi-oxd (approved):
Looks good!

src/scheduler/email-alerts/scripts/advisoryEmailMetadata.js
```diff
@@ -0,0 +1,106 @@
...
+            value: formatMetadataDate(advisory.advisoryDate),
+          }
+        : null;
+    case METADATA_FIELDS.UPDATED_DATE:
+      return advisory.updatedDate
+        ? {
+            label: METADATA_LABELS[fieldId],
+            value: formatMetadataDate(advisory.updatedDate),
+          }
+        : null;
```
ayumi-oxd, line 59: Sorry it's not a question/comment about your code, but do you happen to remember how `updatedDate` is used?

molund: @ayumi-oxd `updatedDate` is entered manually by users and has no real logic tied to it.  It's one of the options when users select which date to display for the advisory.

src/cms/src/api/public-advisory/services/scheduling.js
```diff
@@ -5,6 +5,7 @@
  */
 
 const { queueAdvisoryEmail } = require("../../../helpers/taskQueue.js");
+const { METADATA_FIELDS } = require("../../../helpers/advisoryEmailMetadata.js");
```
molund, line 8: src/scheduler/email-alerts/scripts/sendAdvisoryEmails.js calls a function called `getAdvisoryInfo()` that gets the latest publicAdvisoryAudit.  Couldn't we get a lot of this data from there instead of passing it into the queued tasks?

duncan-oxd: To clarify, this METADATA_FIELDS and the stuff we're passing through is just a list of the field names to include (since each email has a different list of fields to show in the bottom section). The actual values for those fields does come from the same advisory info object 👍

molund (approved):

PR #1871, opened 2026-06-17
CMS-1796: Add endDateReached email and trigger function

duncan-oxd:
### Jira Ticket:
CMS-1796

### Description:

This adds a new email trigger function for "End date reached" - using the same 3-minute window as the other checks (like "expiringSoon")

I also added a few guards to the email template to prevent rendering empty paragraph tags, and added the Rec Resources (name and ID) among the "Affected areas" part of the email, because that was missed in the requirements for the previous email tickets.

### DIFF
--- a/src/scheduler/email-alerts/scripts/sendAdvisoryEmails.js
+++ b/src/scheduler/email-alerts/scripts/sendAdvisoryEmails.js
@@ -231,6 +231,7 @@ const getAdvisoryInfo = async function (advisoryNumber) {
           fields: ["siteName", "slug"],
           populate: { protectedArea: { fields: "slug" } },
         },
+        recreationResources: { fields: ["resourceName", "recResourceId"] },
         standardMessages: { fields: ["description"] },
         urgency: { fields: ["urgency"] },
       },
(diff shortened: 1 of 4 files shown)


duncan-oxd:
The RST Rec Resource prints out as plain text with no link at the moment, similar to how it does on the Summary page on the frontend:
```html
  <p style="margin-top: 4px">
    
    <span style="margin-right:4px;">Juliet Creek Cabin (REC6959)</span>
    
  </p>
```

ayumi-oxd (approved):
Looks good!

src/scheduler/email-alerts/scripts/sendAdvisoryEmails.js
```diff
@@ -231,6 +231,7 @@ const getAdvisoryInfo = async function (advisoryNumber) {
           fields: ["siteName", "slug"],
           populate: { protectedArea: { fields: "slug" } },
         },
+        recreationResources: { fields: ["resourceName", "recResourceId"] },
```
molund, line 234: good catch

PR #1873, opened 2026-06-19
CMS-1813: Strip querystring from YouTube url before calling noembed.com

molund:
### Jira Ticket:
CMS-1813

### Description:
This change fixes an issue with the Berg Lake Trail video where the YouTube URL had a start time querystring and it was breaking the lookup on noembed.com. 

e.g. `https://www.youtube.com/embed/-_YsZHp4IAk?start=165` => `https://noembed.com/embed?dataType=json&url=https://www.youtube.com/watch?v=-_YsZHp4IAk`

I discovered while working on this that cheerio is actually supposed to be a server-side nodeJs library and we're using it on client side in a `useEffect()`. That might explain why it was trying to bundle nodeJs libraries into our webpack bundle.


ayumi-oxd (approved):
Wow, thank you for looking into that!

PR #1880, opened 2026-06-26
CMS-1797: Check isDisplayed in RST resource link code

duncan-oxd:
### Jira Ticket:
CMS-1797

### Description:

Updating the ACT notification email's RST link logic to match the frontend after clarifying that we need to consider `isDisplayed` in our meeting this morning.

Now the link will be left "null" and not used in the email template unless isDisplayed is true.


molund:
I'm going to merge this, but note that in the template the link and the recResourceID are two different concerns now. If there is no link we might still want to show the ID. The template will only show the name if the link is missing.

molund (approved):

PR #1883, opened 2026-07-03
CMS-1555: Change reservations call centre content display logic

ayumi-oxd:
### Jira Ticket:
CMS-1555

### Description:
- Changed the reservations call centre content display logic
- Used `inReservationSystem` instead of other reservation boolean fields


duncan-oxd (approved):
Looks like a nice simplification of the logic!

PR #1887, opened 2026-07-07
CMS-1826: Call ORCA CRUD API from scheduler

molund:
### Jira Ticket:
CMS-1826

### Description:
Implement integration to sync advisory changes to RecSpace's CRUD API. Processes queued tasks created by #1882, authenticating via Keycloak with provided credentials to obtain bearer tokens. Supports POST requests for scheduled/published advisories and DELETE requests for removed resources or unpublished statuses. Handles 404 errors gracefully and logs all operations for debugging purposes.


src/scheduler/server.js
```diff
@@ -76,6 +79,7 @@ const { dootPublish } = require("./doot/scripts/publish");
       await triggerAdvisories();
 
       await dootPublish();
+      await publishToRecSpace();
```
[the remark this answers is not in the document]
molund: Removed import alias

infrastructure/helm/deployment/templates/secrets/etl-secret.yaml
```diff
@@ -15,4 +15,5 @@ metadata:
 stringData:
   STRAPI_API_TOKEN: ""
   PARK_NAMES_API_KEY: ""
+  REC_SPACE_KEYCLOAK_CLIENT_SECRET: ""
```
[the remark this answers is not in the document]
molund: I added it to all 5 environments manually

src/scheduler/shared/taskQueue.js
```diff
@@ -3,7 +3,7 @@ const { getLogger } = require("./logging");
 const qs = require("qs");
 
 const readQueue = async function (actionName, options) {
-  const sort = options?.descending ? "numericData:DESC" : "numericData";
+  let sort = [options?.descending ? "numericData:DESC" : "numericData", "id"];
```
duncan-oxd: Would this still work as a `const`?

src/scheduler/recspace/scripts/publishAdvisories.js
```diff
@@ -0,0 +1,116 @@
...
+      const afterStatus = afterJsonData?.advisory_status || null;
+
+      // if the after status is "Scheduled" or "Published" then we need to send a POST request
+      // to the API for each recResourceId in the after payload
+      // Note: POST endpoint performs upsert, so PUT is unnecessary
+      if (afterStatus === "Scheduled" || afterStatus === "Published") {
+        for (const recResourceId of afterJsonData?.rec_resource_ids || []) {
+          const payload = { ...afterJsonData };
+          delete payload.revision_number;
+          delete payload.rec_resource_ids;
```
duncan-oxd: This tripped me up for a moment 😅 We could probably define something like `basePayload` outside of the loop and just add the ID inside the loop. Minor nitpick just because I got confused!

duncan-oxd (approved):
Looks good! I had one small note to consider. Do you know if there's any kind of rate limiter on the API's side that might cause grief if we're sending a bunch of sequential requests?

molund:
> Looks good! I had one small note to consider. Do you know if there's any kind of rate limiter on the API's side that might cause grief if we're sending a bunch of sequential requests?

I'm not sure, but it's probably okay because most advisories will only have a single rec resource, and only one advisory gets updated at at time.

src/scheduler/shared/taskQueue.js
molund: Good catch. I was trying something different when it got changed to `let`

src/scheduler/recspace/scripts/publishAdvisories.js
molund: Thanks

PR #1888, opened 2026-07-08
CMS-1826: RST API Helm refactoring

molund:
### Jira Ticket:
CMS-1826

### Description:
Combined ETL settings with scheduler settings since they both use the RST API. 

I could have standardized the environment variable names and URL path too, but I opted against that because it would change `.env` and `.env.example` in the etl and scheduler projects requiring everyone to update their local dev environments.

### DIFF
--- a/infrastructure/helm/deployment/values-alpha-test.yaml
+++ b/infrastructure/helm/deployment/values-alpha-test.yaml
@@ -2,6 +2,7 @@ cluster:
   ssoAuthUrl: https://test.loginproxy.gov.bc.ca/auth
   publicUrl: https://alpha-test.bcparks.ca
   staffPortalUrl: https://alpha-test-staff.bcparks.ca
+  recSpaceApiUrl: https://dpeu7dmgz4j6y.cloudfront.net
 
 images:
   strapi:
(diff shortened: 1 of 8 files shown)


infrastructure/helm/deployment/values.yaml
```diff
@@ -4,6 +4,7 @@ cluster:
   ssoRealm: bcparks-service-transformation
   ssoClientId: staff-portal
   staffPortalUrl: https://dev-staff.bcparks.ca
+  recSpaceApiUrl: https://dv5oa6g5xh0tq.cloudfront.net
```
molund, line 7: Used dev as the default (was previously test)

infrastructure/helm/deployment/values-alpha-test.yaml
```diff
@@ -2,6 +2,7 @@ cluster:
   ssoAuthUrl: https://test.loginproxy.gov.bc.ca/auth
   publicUrl: https://alpha-test.bcparks.ca
   staffPortalUrl: https://alpha-test-staff.bcparks.ca
+  recSpaceApiUrl: https://dpeu7dmgz4j6y.cloudfront.net
```
molund, line 5: This URL is from the Confluence page: 
https://apps.nrs.gov.bc.ca/int/confluence/display/BCPRS/Testing+Act+Integration+CUD+Endpoints+via+Swagger+UI?ct=1783355437035

There are actually 5 RecSpace URLs right now. It might go back down to 3.

duncan-oxd (approved):
LGTM! Be sure to give QA a heads up before you apply these changes, because they'll wipe out any runtime customization for the scheduler email recipients/whitelist 👍

PR #1889, opened 2026-07-09
CMS-1916: Fix gatsby build error

ayumi-oxd:
### Jira Ticket:
CMS-1916

### Description:
- I'm not entirely sure why it got broken on PROD
  - The build error happens only on PROD, doesn't happen on DEV or TEST
  - The build on PROD doesn't create the cache: https://github.com/bcgov/bcparks.ca/actions/caches
- Added safe guards to safely access nested properties using optional chaining and default values, preventing crashes when data is missing
- Improved utility functions such as `getFeatureDates` and `getParkDates` to handle undefined or empty arrays, ensuring safer date formatting and display logic
- Modified the Gatsby schema in `gatsby-node.js` to add new types (`STRAPI_PROTECTED_AREA_DESCRIPTION_TEXTNODE`, `STRAPI_PROTECTED_AREADescription`) and updated the `STRAPI_PROTECTED_AREA` type to link to the new description structure, supporting richer and more flexible data modeling
- Fixed potential errors in data transformation logic (e.g., in `active-advisories.js`) by adding checks for missing IDs and defaulting to empty arrays, ensuring that data hashes are built only with valid entries

### DIFF
--- a/src/gatsby/gatsby-node.js
+++ b/src/gatsby/gatsby-node.js
@@ -224,7 +224,16 @@ exports.createSchemaCustomization = ({ actions }) => {
     updatedAt: Date
   }
 
+  type STRAPI_PROTECTED_AREA_DESCRIPTION_TEXTNODE implements Node @dontInfer {
+    description: String
+  }
+
+  type STRAPI_PROTECTED_AREADescription {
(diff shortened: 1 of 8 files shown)


molund (approved):
I was able to reproduce the error by attaching my local Gatsby to the prod cms.  
```
npm run clean
npm run build
```
With your changes the error is fixed.  Thanks for taking care for of this.

PR #1890, opened 2026-07-10
CMS-1889: Fix advisory loading bug

ayumi-oxd:
### Jira Ticket:
CMS-1889

### Description:
- The bug happened because the search endpoint ignored offset pagination from the frontend (`pagination[start]`/`[pagination[limit]`) and only handled page-based params, so every “Load more” call effectively returned page 1 again from `search.js`

### DIFF
--- a/src/cms/src/api/public-advisory/services/search.js
+++ b/src/cms/src/api/public-advisory/services/search.js
@@ -9,12 +9,27 @@ module.exports = ({ strapi }) => ({
   search: async (query) => {
     query = buildQuery(query);
 
+    const paginationLimit = Number(query.pagination?.limit);
+    const paginationPageSize = Number(query.pagination?.pageSize);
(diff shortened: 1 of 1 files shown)


src/cms/src/api/public-advisory/services/search.js
```diff
@@ -9,12 +9,47 @@ module.exports = ({ strapi }) => ({
...
+        return undefined;
+      }
+
+      if (typeof value === "string" && value.trim() === "") {
+        return undefined;
+      }
+
+      const numericValue = Number(value);
+      return Number.isInteger(numericValue) ? numericValue : undefined;
+    };
```
duncan-oxd, line 23: Oof, we need typescript 😆

duncan-oxd (approved):
Looks good! Hopefully you have time to merge this and send to QA today. If not, I can move it over next week while you're away 👍

PR #1906, opened 2026-07-30
CMS-946: Use update date for advisories sorting first

ayumi-oxd:
### Jira Ticket:
CMS-946

### Description:
- The advisories should be sorted by the updated date even if the display date is not using updated date
- If the update date is null for an advisory, then it should use the advisory date


molund (approved):
Clever

PR #1908, opened 2026-08-04
CMS-946: Fix advisories sorting by dates

ayumi-oxd:
### Jira Ticket:
CMS-946

### Description:
- Previously, `eventType.precedence` was the last sort criterion and was compared after a millisecond-precision date, meaning two advisories would almost never tie on date and event type sorting would never take effect. 
- Now, the date is compared at day precision only (time of day is ignored), so advisories set to the same calendar day compare as equal and get sorted by event type as intended.

### DIFF
--- a/src/gatsby/src/components/park/advisoryDetails.js
+++ b/src/gatsby/src/components/park/advisoryDetails.js
@@ -25,6 +25,16 @@ const getTimestamp = isoDate => {
   return isoDate ? parseJSON(isoDate).getTime() : null
 }
 
+// Returns a day-precision timestamp for sorting: 
+// Uses updatedDate when available, falls back to advisoryDate.
+// Same-day advisories compare as equal, allowing eventType.precedence to act as the tiebreaker.
+const getEffectiveDayTimestamp = advisory => {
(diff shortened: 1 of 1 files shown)


src/gatsby/src/components/park/advisoryDetails.js
```diff
@@ -21,8 +21,14 @@ const formatDate = isoDate => {
...
-const getTimestamp = isoDate => {
-  return isoDate ? parseJSON(isoDate).getTime() : null
+// Returns a day-precision timestamp for sorting: 
+// Uses updatedDate when available, falls back to advisoryDate.
+// Same-day advisories compare as equal, allowing eventType.precedence to act as the tiebreaker.
+const getEffectiveDayTimestamp = advisory => {
+  const d = advisory.updatedDate ?? advisory.advisoryDate
+  if (!d) return null
+  const date = parseJSON(d)
+  return new Date(date.getFullYear(), date.getMonth(), date.getDate()).getTime()
```
molund: date-fns has a `startOfDay` helper for this as well

```
return startOfDay(parseJSON(d)).getTime()
```

molund (approved):

PR #1919, opened 2026-08-19
CMS-1987: Adjust email queue de-dupe logic to check subject

duncan-oxd:
### Jira Ticket:
CMS-1987

### Description:

Some tweaks to the email queue system, now that there are more emails and more recipients involved:

- Instead of de-duplicating emails just based on the advisoryNumber, now consider the advisoryNumber and subject. This allows multiple emails to be dispatched for the same event without being discarded as duplicates
- Adjusted the email throttling that discarded more than one email per advisory per queue, as well as the 10-minute throttle. now they both check the subject to make sure the email really is a duplicate before discarding it.
- For testing and troubleshooting, I also added the email addresses of the recipients to the files that get generated if you run `node manage.js emailtest`. It shows up in an HTML comment so you can see it if you view source.
- I also added the documentId to the name of the files generated from `emailtest`, since an advisory might have more than one email in the queue at the same time.

### DIFF
--- a/src/cms/src/helpers/taskQueue.js
+++ b/src/cms/src/helpers/taskQueue.js
@@ -102,6 +102,7 @@ module.exports = {
           action: "email advisory",
           numericData: advisoryNumber,
         },
+        fields: ["jsonData"],
       });
 
     // Check if any existing queued tasks for this advisory have the same subject
--- a/src/cms/src/helpers/taskQueue.js
+++ b/src/cms/src/helpers/taskQueue.js
@@ -95,15 +95,20 @@ module.exports = {
     if (!subject || !title || !advisoryNumber) {
       return;
     }
-    const exists =
-      (
-        await strapi.documents("api::queued-task.queued-task").findMany({
-          filters: {
-            action: "email advisory",
-            numericData: advisoryNumber,
-          },
-        })
-      ).length > 0;
+    const existingTasks = await strapi
+      .documents("api::queued-task.queued-task")
+      .findMany({
+        filters: {
+          action: "email advisory",
+          numericData: advisoryNumber,
+        },
+      });
+
+    // Check if any existing queued tasks for this advisory have the same subject
+    const exists = existingTasks.some(
+      (task) => task?.jsonData?.subject === subject,
+    );
+
     if (!exists) {
       strapi.log.info(
         `queued advisoryNumber ${advisoryNumber} for "${subject}" notification`,
--- a/src/scheduler/email-alerts/scripts/sendAdvisoryEmails.js
+++ b/src/scheduler/email-alerts/scripts/sendAdvisoryEmails.js
@@ -17,7 +17,7 @@ const { buildEmailMetadata } = require("./advisoryEmailMetadata");
  * Sends queued emails
  */
 exports.sendAdvisoryEmails = async function (recentAdvisoryEmails) {
-  const THROTTLE_MINUTES = 10; // min. time before sending 2 emails about 1 advisory
+  const THROTTLE_MINUTES = 10; // min. time before sending duplicate emails for an advisory+subject
   let queue;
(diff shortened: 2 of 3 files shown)


src/scheduler/email-alerts/templates/public-advisory.ejs
```diff
@@ -6,6 +6,10 @@
   <title><%= title %></title>
 </head>
 
+<% if (allRecipients && allRecipients.length > 0) { %>
+<!-- Email recipients: <%= allRecipients.join(", ") %> -->
+<% } %>
```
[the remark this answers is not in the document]
duncan-oxd: The current version is safe for all expected uses: The email recipients only get added to `allRecipients` in the `emailtest` case, so there's no danger of this happening when actual emails are sending / in prod.

If the `allRecipients` variable is ever used for any other purpose, we can check for an explicit test flag in the template to harden this logic.

src/scheduler/email-alerts/scripts/sendAdvisoryEmails.js
```diff
@@ -208,6 +223,10 @@ exports.sendAdvisoryEmails = async function (recentAdvisoryEmails) {
           );
         }
       }
+    } else {
+      logger.info(
+        `Throttled advisory email ${advisoryNumber} with subject "${throttleSubject}". Removing duplicate queue item.`,
+      );
     }
```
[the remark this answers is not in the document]
duncan-oxd: Updated that log line to be more explicit about what it's doing: "skipped" instead of "throttled", and took out the part about the queue.

src/cms/src/helpers/taskQueue.js
```diff
@@ -95,15 +95,20 @@ module.exports = {
...
-        })
-      ).length > 0;
+    const existingTasks = await strapi
+      .documents("api::queued-task.queued-task")
+      .findMany({
+        filters: {
+          action: "email advisory",
+          numericData: advisoryNumber,
+        },
+      });
```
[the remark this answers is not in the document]
duncan-oxd: There are only two fields in that model, and jsonData is the big one anyway, but sure :P

src/cms/notes.txt
```diff
@@ -0,0 +1,22 @@
+need a way to test / isolate. got 3 branches
```
molund: Was this file included by accident?

src/scheduler/email-alerts/scripts/sendAdvisoryEmails.js
```diff
@@ -38,11 +38,17 @@ exports.sendAdvisoryEmails = async function (recentAdvisoryEmails) {
 
   for (const message of queue) {
     const advisoryNumber = message?.numericData;
+    const throttleSubject = message?.jsonData?.subject || "";
+
     if (
-      !recentAdvisoryEmails.find((e) => e.advisoryNumber === advisoryNumber)
+      !recentAdvisoryEmails.find(
```
molund, line 44: e is usually shorthand for error. Maybe just use the word "email"

src/cms/notes.txt
duncan-oxd: 😆 Yep, this was a stashed file that got committed by accident. Oops!
These were my notes about the 3.0.9 deploy issues that I was going over in the meeting yesterday. I'll delete this shortly

molund (approved):

src/scheduler/email-alerts/scripts/sendAdvisoryEmails.js
duncan-oxd: I try to never use single-character var names because the tradeoff for brevity vs. clarity usually does more harm than good when I'm reading the file later. This branch just changed the indent on that existing line, but I've now updated two spots to use "email" as the var name in the arrow function 👍

PR #1922, opened 2026-08-19
NOBUG: Revert sort order fix in gatsby config

duncan-oxd:
### Jira Ticket:
None

### Description:

_One of several branches I'm pushing for review/discussion._ 

This branch reverts the query sort order changes I made for "v3.0.4.9" - if we wanted to deploy it somewhere to test and see if we can isolate whether or not that's the change that fixed the build issues. I recognize that we don't have any environments to spare right now with all the UAT going on, though.

### DIFF
--- a/src/gatsby/gatsby-config.js
+++ b/src/gatsby/gatsby-config.js
@@ -83,7 +83,6 @@ module.exports = {
             singularName: "protected-area",
             queryParams: {
               fields: "*",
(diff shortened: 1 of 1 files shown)


ayumi-oxd (approved):

duncan-oxd:
We won't be deploying this. We investigated the incident quite a bit and are pretty confident about the cause at this point.

PR #1928, opened 2026-08-26
CMS-2000: Add deterministic sort ordering to multi-page queries

duncan-oxd:
### Jira Ticket:
CMS-2000

### Description:

Did a review through the repo for paginated data fetches that didn't have a fully deterministic sort order and added something like ID/ORCS where appropriate.

Tested by re-building the ElasticSearch index locally and clicking "Load more" on various park searches, compared with the behavior on the production site.

I also updated a few scripts and tested them with a "dry run" by replacing the db writes with console logs locally. It all looks good to me, but I'm looking for feedback! 

I have a few similar tweaks to make for the SP repo as well. 

TBD how we want to deploy this once it's tested. It could be a hotfix or it could wait until Oct. I haven't added any version tags yet.

### DIFF
--- a/src/cms/src/api/protected-area/services/search.js
+++ b/src/cms/src/api/protected-area/services/search.js
@@ -114,7 +114,7 @@ module.exports = ({ strapi }) => ({
 
     let sortOrder;
     if (isNaN(latitude) || isNaN(longitude)) {
-      sortOrder = ["_score", "nameLowerCase.keyword"];
+      sortOrder = ["_score", "nameLowerCase.keyword", "orcs"];
     } else {
       sortOrder = [
         {
@@ -127,6 +127,10 @@ module.exports = ({ strapi }) => ({
             ignore_unmapped: true,
           },
         },
+        // if two parks are the same distance away, sort by score and then by keyword
+        { _score: { order: "desc" } },
+        { "nameLowerCase.keyword": { order: "asc" } },
+        { orcs: { order: "asc" } },
       ];
     }
 
--- a/src/etl/scripts/rst-resources.js
+++ b/src/etl/scripts/rst-resources.js
@@ -221,15 +221,12 @@ async function fetchRSTResources() {
   const rstSummaryUrl = `${process.env.RST_API}/recreation-resource/summary`;
 
   const rstResources = [];
-  const { data } = await axios.get(`${rstSummaryUrl}?page=1&sort=id`, rstAxiosConfig);
+  const { data } = await axios.get(`${rstSummaryUrl}?page=1`, rstAxiosConfig);
   const totalPages = data.totalPages;
   rstResources.push(...data.data);
 
   for (let page = 2; page <= totalPages; page++) {
-    const { data } = await axios.get(
-      `${rstSummaryUrl}?page=${page}&sort=id`,
-      rstAxiosConfig,
-    );
+    const { data } = await axios.get(`${rstSummaryUrl}?page=${page}`, rstAxiosConfig);
     rstResources.push(...data.data);
   }
   return rstResources;
--- a/src/etl/scripts/rst-resources.js
+++ b/src/etl/scripts/rst-resources.js
@@ -221,12 +221,15 @@ async function fetchRSTResources() {
   const rstSummaryUrl = `${process.env.RST_API}/recreation-resource/summary`;
 
   const rstResources = [];
-  const { data } = await axios.get(`${rstSummaryUrl}?page=1`, rstAxiosConfig);
+  const { data } = await axios.get(`${rstSummaryUrl}?page=1&sort=id`, rstAxiosConfig);
   const totalPages = data.totalPages;
   rstResources.push(...data.data);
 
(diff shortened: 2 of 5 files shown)


src/cms/src/api/protected-area/services/search.js
```diff
@@ -114,7 +114,7 @@ module.exports = ({ strapi }) => ({
 
     let sortOrder;
     if (isNaN(latitude) || isNaN(longitude)) {
-      sortOrder = ["_score", "nameLowerCase.keyword"];
+      sortOrder = ["_score", "nameLowerCase.keyword", "orcs"];
```
molund, line 117: Name should already be deterministic. No two parks have the same name.

src/etl/scripts/rst-resources.js
```diff
@@ -221,12 +221,15 @@ async function fetchRSTResources() {
...
 
   const rstResources = [];
-  const { data } = await axios.get(`${rstSummaryUrl}?page=1`, rstAxiosConfig);
+  const { data } = await axios.get(`${rstSummaryUrl}?page=1&sort=id`, rstAxiosConfig);
   const totalPages = data.totalPages;
   rstResources.push(...data.data);
 
   for (let page = 2; page <= totalPages; page++) {
-    const { data } = await axios.get(`${rstSummaryUrl}?page=${page}`, rstAxiosConfig);
+    const { data } = await axios.get(
```
molund: This is actually calling the RST API.  It doesn't have a field called "id", and it looks like it's already sorted by rec_resource_id

https://dv5oa6g5xh0tq.cloudfront.net/api/v1/recreation-resource/summary

src/scheduler/temporary-scripts/rst-import/import-rec-resources.js
```diff
@@ -215,13 +215,13 @@ async function fetchRSTResources() {
   console.log(`Fetching recreation resources from RST API: ${rstSummaryUrl}`);
 
   const rstResources = [];
-  const { data } = await axios.get(`${rstSummaryUrl}?page=1`, rstAxiosConfig);
+  const { data } = await axios.get(`${rstSummaryUrl}?page=1&sort=id`, rstAxiosConfig);
```
molund: As above, this is the RST API

src/cms/src/api/protected-area/services/search.js
duncan-oxd: There are quite a few spots that _should_ always be unique for business logic but they don't have a particular unique constraint in the DB, so I added a tie-breaker anyway, but you're right 👍

src/etl/scripts/rst-resources.js
duncan-oxd: Ah! I was indiscriminately searching the whole repo for axios calls and I got carried away. I'll revert the RST ones.

src/etl/scripts/rst-resources.js
```diff
@@ -235,6 +235,7 @@ async function fetchRSTResources() {
 // Fetches all recreation resources from Strapi, handling pagination
 async function fetchStrapiResources(strapiResourcesUrl) {
   const queryParams = {
+    sort: ["id:asc"],
```
ayumi-oxd, line 238: I think it'll still work but Strapi v5 uses `documentId` as the primary key.

src/scheduler/temporary-scripts/rst-import/import-closures.js
```diff
@@ -64,7 +64,10 @@ const loadData = async function () {
     for (let page = 1, pageCount = 1; page <= pageCount; page++) {
       const { data } = await axios.get(url, {
         headers: httpReqHeaders,
-        params: { pagination: { page, pageSize: 1000 } },
+        params: {
+          sort: ["id:asc"],
```
ayumi-oxd, line 68: The same as above.

src/scheduler/temporary-scripts/rst-import/import-rec-resources.js
```diff
@@ -232,6 +232,7 @@ async function fetchRSTResources() {
 // Fetches all recreation resources from Strapi, handling pagination
 async function fetchStrapiResources(strapiResourcesUrl) {
   const queryParams = {
+    sort: ["id:asc"],
```
ayumi-oxd, line 235: The same as above.

src/etl/scripts/rst-resources.js
duncan-oxd: I checked the schema just now, and it looks like the `id` integer field is the primary key on all of Strapi's DB tables behind the scenes, but documentId is the public-facing identifier in the CMS, for whatever reason.

The really interesting part is that documentId (`document_id` in the db) **does not have a UNIQUE constraint**. For that reason, I'll advise sticking with `id` for these sorting tie-breakers.

molund: Confirming id should still work in Strapi 5. This was the exact pattern used for the hotfix in July

molund (approved):

PR #1929, opened 2026-08-26
CMS-1955: Add pruning functionality to Gatsby rollback images

molund:
### Jira Ticket:
CMS-1955

### Description

This PR adds retention management for rollback images generated by the Publish Gatsby GitHub Action.

A new retention policy has been implemented for the `public-main` ImageStream, allowing historical images to be retained for rollback purposes while preventing unbounded growth of image tags in OpenShift.

### Retention Rules

For `public-main:rollbackYYYYMMDDTHHMM` tags:

- Days 0-3: Keep every image
- Days 4-14: Keep only the first image created each day
- Older than 14 days: Delete
- Always retain a minimum number of rollback builds

The `imagetag-pruner` has been expanded to support these Gatsby-specific retention rules while preserving the existing pruning behavior for build artifact images.

### Additional Changes

This PR also incorporates the fixes and cleanup from https://github.com/bcgov/bcparks-staff-portal/pull/727 to keep the imagetag-pruner implementation synchronized across repositories:

- SHA tags are now only included on `main` branch builds.  The tags on `alpha` builds were never used and they cause the main branch tags to expire more quickly.
- Updated GitHub Actions to generate 8-character `SHORT_SHA` values, aligning image tagging conventions with the companion repository and ensuring pruning rules remain consistent.
- Removed an unused `SHORT_SHA` variable from a GitHub Action that was identified during the update.
- Fixed an issue where the pruner incorrectly matched 7-character SHA tags after `SHORT_SHA` was increased from 7 to 8 characters.
- Fixed environment variable typo:
  - `IMAGESTRAMS_TO_CLEAN` → `IMAGESTREAMS_TO_CLEAN`
- Added local debugging support:
  - `.env` support
  - `.env.example`
  - `npm run start:local`
- Added the missing `package-lock.json` file


tools/imagetag-prune/imagePruner.js
```diff
@@ -76,17 +92,49 @@ class ImagePruner {
...
       }
     }
   }
 
-  #shouldCleanImage(name, tag) {
+  #shouldCleanBuildArtifactImage(name, tag) {
+    // Gatsby rollback imagestream is handled separately in #shouldCleanGatsbyRollbackImage
+    if (name === this.gatsbyRollbackImageName) {
+      return false;
+    }
```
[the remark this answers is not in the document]
molund: This imagestream only ever has `latest`, `test`, `prod` and `rollbackYYYYMMDDTHHMM` tags. There are no short SHA or release tags.

tools/imagetag-prune/Dockerfile
```diff
@@ -1,8 +1,10 @@
-FROM node:22-slim
+FROM --platform=linux/amd64 node:24-slim
```
ayumi-oxd, line 1: Was v24 compatible with Strapi v5?

molund: It doesn't look like it. 

I just tried with the version of Strapi in the alpha branch on https://github.com/bcgov/bcparks.ca and it still has some warning.

But this is for the image tag pruner: https://github.com/bcgov/bcparks-staff-portal/tree/main/tools/imagetag-prune

ayumi-oxd (approved):

.github/workflows/build-dev.yaml
```diff
@@ -21,7 +21,12 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Set env
-        run: echo "SHORT_SHA=$(git rev-parse --short HEAD)" >> $GITHUB_ENV
+        run: |
+          echo "SHORT_SHA=$(git rev-parse --short=8 HEAD)" >> $GITHUB_ENV
+          echo "IMAGE_TAGS=latest" >> $GITHUB_ENV
+          if [ "${GITHUB_REF_NAME}" = "main" ]; then
+            echo "IMAGE_TAGS=latest ${SHORT_SHA}" >> $GITHUB_ENV
+          fi
```
duncan-oxd: If these lines both append to a temporary environment file that github uses to set variables, we could change this into an if/else block to avoid setting and then re-setting the `IMAGE_TAGS` variable

duncan-oxd (changes requested):
Nice! We'll be glad to have the expanded retention policy. I had one note about setting the variable but the rest looks good to go.

.github/workflows/build-dev.yaml
molund: Updated the code

PR #1931, opened 2026-09-08
CMS-1543: Update park access status endpoint

ayumi-oxd:
### Jira Ticket:
CMS-1543

### Description:
- Updated the park access status endpoint
  - Brought the calculation to display "Seasonal restrictions" from FE to the controller
  - You can check it locally at http://localhost:1337/api/park-access-statuses

### DIFF
--- a/src/cms/src/api/protected-area/custom/protected-area-status.js
+++ b/src/cms/src/api/protected-area/custom/protected-area-status.js
@@ -1,5 +1,7 @@
 "use strict";
 const _ = require("lodash");
+const format = require("date-fns/format");
+const utcToZonedTime = require("date-fns-tz/utcToZonedTime");
 
 const boolToYN = (boolVar) => {
   return boolVar ? "Y" : "N";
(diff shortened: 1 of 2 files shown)


molund (approved):
Nice code refactoring and reuse.  

This is the kind of change that makes me wish we had already implemented Turborepo (CMS-1620)

I think we can probaly remove the DO NOT MERGE tag from this one and merge it, since it's going to alpha and not main. But maybe check with QA first.

src/cms/src/api/protected-area/custom/protected-area-status.js
```diff
@@ -11,6 +13,67 @@ const getHasCampfiresFacility = (parkFacilities) => {
   );
 };
 
+// dateTypeId values from park-date-type
+const GATE_DATE_TYPE_ID = 1;
+const OPERATION_DATE_TYPE_ID = 6;
```
molund, line 18: These will be renamed in CMS-1342

PR #1932, opened 2026-09-09
CMS-2004: Move email whitelist env var into a secret

duncan-oxd:
### Jira Ticket:
CMS-2004

### Description:

Converting our recently-added `EMAIL_RECIPIENT_WHITELIST` env var to a secret, to prevent it being exposed in the source code or reset when helm upgrades are applied.

I made a generic secret for any scheduler-related secrets, following the patterns for the other secrets. Default value is "" (we'll leave it blank on prod) and we'll manually update it on each non-prod environment to have all of our email addresses.

I haven't applied this on any environments yet, but I verified the rendered template with e.g.:
```sh
helm template alpha . -f values-alpha-dev.yaml --show-only templates/secrets/scheduler-secret.yaml
# or the entire template
helm template alpha . -f values-alpha-dev.yaml 
```

Tagging with "cms-october" and "act-cms" as it's an infrastructure change that doesn't require testing, so it could be included in the ACT release if it's approved by then.


infrastructure/helm/deployment/templates/secrets/scheduler-secret.yaml
```diff
@@ -0,0 +1,17 @@
...
+    component: {{ .Values.scheduler.componentName }}
+    release: {{ .Release.Name }}
+    chart: {{ .Chart.Name }}
+    heritage: {{ .Release.Service }}
+  annotations:
+    "helm.sh/hook": "pre-install,pre-upgrade"
+    "helm.sh/hook-delete-policy": "before-hook-creation"
+    "helm.sh/hook-weight": "0"
+stringData:
+  EMAIL_RECIPIENT_WHITELIST: ""
```
molund, line 16: maybe we should default it to `cardweb@gov.bc.ca` instead of leaving it blank?

infrastructure/helm/deployment/templates/scheduler/scheduler-deployment.yaml
```diff
@@ -60,7 +60,10 @@ spec:
             - name: EMAIL_RECIPIENT
               value: {{ .Values.scheduler.env.emailRecipient | quote }}
             - name: EMAIL_RECIPIENT_WHITELIST
-              value: {{ .Values.scheduler.env.emailRecipientWhitelist | quote }}
+              valueFrom:
+                secretKeyRef:
+                  name: {{ template "bcparks_scheduler_secret" . }}
+                  key: EMAIL_RECIPIENT_WHITELIST
```
molund, line 66: The scheduler now loads two secrets from etl-secret (STRAPI_API_TOKEN and REC_SPACE_KEYCLOAK_CLIENT_SECRET) and one secret from scheduler-secret.

This isn't a blocker, but having scheduler-related secrets split across multiple secret resources may be a source of confusion for future maintenance and troubleshooting. 

Maybe the etl secret needs to be renamed something like `main-automation-secret`/`alpha-automation-secret`

infrastructure/helm/deployment/templates/secrets/scheduler-secret.yaml
duncan-oxd: An empty string is valid (for prod, at least) where `sendAdvisoryEmails` splits checks the length of the list. If we want to have a default, I'd sooner have `change_me@example.com` than something that could someday become incorrect or cause confusion. 

In any case, we'd need to go in and manually update them all once the secrets are created (except prod, which we'd either clear or leave blank).

Any preference?

infrastructure/helm/deployment/templates/scheduler/scheduler-deployment.yaml
duncan-oxd: Would it make sense to move those two other values into this secret and have it be the one and only "scheduler secret"?

infrastructure/helm/deployment/templates/secrets/scheduler-secret.yaml
molund: I think blank is fine. Like you said, PROD doesn't matter. TEST is the only environment set to cardweb. All the others are set to mike & manuji.

infrastructure/helm/deployment/templates/scheduler/scheduler-deployment.yaml
molund: Perhaps the best practice would be to have separate Strapi tokens for ETL and the scheduler? Technically, we should be using the least privileged tokens for both.

#### Scheduler secrets
- Scheduler Strapi token
- REC_SPACE_KEYCLOAK_CLIENT_SECRET
- EMAIL_RECIPIENT_WHITELIST

#### ETL secrets
- ETL Strapi token
- PARK_NAMES_API_KEY

molund:
Let's include this in the ACT release. I'll remove the cms-october tag, but I can't change the base without removing the stack.

infrastructure/helm/deployment/templates/scheduler/scheduler-deployment.yaml
duncan-oxd: I like that idea, I'll make a ticket to follow-up with those changes, because they should probably happen after the cleanup in #1933

duncan-oxd: Created CMS-2037 to "Reorganize Scheduler/ETL OpenShift secrets" in Sprint 26

molund (approved):

PR #1935, opened 2026-09-09
CMS-1926: Add delete functionality to doot publish task

ayumi-oxd:
### Jira Ticket:
CMS-1926

### Description:
- Added the delete functionality to DOOT publish task
  - When blank dates were published from DOOT, the same date ranges will be deleted if it's available

### DIFF
--- a/src/scheduler/doot/scripts/publish.js
+++ b/src/scheduler/doot/scripts/publish.js
@@ -292,8 +292,24 @@ exports.dootPublish = async function () {
                   );
                 }
               }
-              // create the date ranges
+              // create/update/delete the date ranges
               for (const dootDateRange of item.dateRanges) {
+                const existingDocId =
+                  dootDateRange.id != null ? existingBySourceId.get(dootDateRange.id) : undefined;
+
+                // a blank startDate/endDate means the date range was intentionally removed in DOOT, 
+                // so the matching Strapi record should be deleted rather than saved with
+                // blank required fields (which would fail validation and abort the whole message).
+                const isBlankDateRange =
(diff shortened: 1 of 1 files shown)


src/scheduler/doot/scripts/publish.js
```diff
@@ -292,8 +292,24 @@ exports.dootPublish = async function () {
                   );
                 }
               }
-              // create the date ranges
+              // create/update/delete the date ranges
               for (const dootDateRange of item.dateRanges) {
+                const existingDocId =
+                  dootDateRange.id != null ? existingBySourceId.get(dootDateRange.id) : undefined;
```
molund, line 298: ParkDates.sourceDateRangeId is a new field and there is no practical way to backfill it. Most 2026 and earlier records therefore have a NULL value.

Once 2027 dates are published from DOOT, newly generated records will populate this field. The problem will resolve itself over time, but I wanted to add some context on why the values are currently missing.

src/scheduler/doot/scripts/publish.js
```diff
@@ -236,7 +236,14 @@ exports.dootPublish = async function () {
             item.dateRanges?.map((dateRange) => dateRange.id).filter((id) => id != null) || [],
           );
 
-          // build a map of sourceDateRangeId -> documentId for existing records
+          // Note: sourceDateRangeId is a newer field added to track which DOOT record created
```
duncan-oxd, line 239: I just mentioned on Slack that this field could cause problems if two environments publish to the same CMS (thinking of the training environment and production environments both connecting to the prod Strapi)

As long as we have a plan to work around that issue, the changes here look good!

duncan-oxd (approved):

PR #1940, opened 2026-09-14
CMS-1543: Add date type to search index return

ayumi-oxd:
### Jira Ticket:
CMS-1543

### Description:
- The accessStatus value differed across the Find a Park page, the individual park page, and the API because the `/api/protected-areas/search` endpoint (used by Find a Park) didn't return `parkDateType`/`dateTypeId` on park feature dates. The field the seasonal-restrictions calculation depends on to identify operating-season dates.


molund (approved):
LGTM. 

Will it ever actually be null now, after the search index is rebuilt?

ayumi-oxd:
> LGTM.
> 
> Will it ever actually be null now, after the search index is rebuilt?

Probably no, the query already filters parkDateType, so once a park is reindexed, parkDateType is guaranteed to be populated. The fallback just covers parks that haven't been reindexed yet since this deployed. The fallback can be removed later.

PR #1941, opened 2026-09-16
CMS-1543: Add status published to getPublishedPublicAdvisories

ayumi-oxd:
### Jira Ticket:
CMS-1543

### Description:
- Issue: A park with three "Open" advisories showed "Open" on the park page but "Seasonal restrictions" on the `/api/park-access-statuses` API.
Screenshot 2026-09-15 at 8 39 34 PM

- Cause: The API's advisory lookup was missing `status: "published"`, so Strapi fetched draft records instead of published ones. The endpoint couldn't match any advisories to the park, so it silently fell back to a hardcoded default instead of the real advisory data, letting the seasonal-restrictions check apply incorrectly.

The reason why `status` is needed even everything is published in public-advisories
- Without a status param on the outer query, Strapi's Document Service defaulted to resolving that relation against the draft version of each protected area, but the actual link table row only points at the published protected-area's internal id. So the relation lookup came up empty, not because the advisory itself was unpublished, but because it was trying to match a related row under the wrong status. The bug was in how the protectedAreas relation got resolved for a status-less query, and status: "published" fixed it by making that resolution consistent with what the link table actually points to.


molund:
I wouldn’t expect status: "published" to be required here, since public-advisory is not D&P-enabled. That said, it still makes sense that it influences related D&P-enabled content.

I did a quick check and didn’t find other examples in the codebase matching:

- non-Draft & Publish base collection
- no explicit status filter
- populate of a Draft & Publish-enabled relation

So this looks like a valid fix, and I don’t see an obvious broader pattern from this quick review.

molund (approved):

PR #1943, opened 2026-09-16
CMS-1543: Include seasonal restriction advisories

ayumi-oxd:
### Jira Ticket:
CMS-1543

### Description:
- Issue: See the bug Adrianne found in the ticket. The park-status endpoint was filtering out advisories with access status `precedence` 120 or higher, which included “Seasonal restrictions.” That meant the status logic never saw the seasonal advisory and fell back to the default Open status.
- Fix: Removed that filter in `protected-area-status.js` so the status calculation can evaluate all published advisories and correctly derive “Seasonal restrictions” when appropriate.


molund (approved):
LGTM. I didn't know seasonal restriction advisories could be created manually

